PR #30, opened 2026-04-22
cloud-env: dispatch shim for SessionStart hooks; integrity-check probe

venpopov:
## Summary

Closes the cloud-side regression introduced in #28: hook commands in `.claude/settings.json` referenced `${CLAUDE_PROJECT_DIR}/scripts/…`, but cloud sessions launch with five `--add-dir` repos so `CLAUDE_PROJECT_DIR` resolves to `/home/user` (multi-root parent), not `/home/user/vade-runtime`. All five `SessionStart` hooks failed with exit 127, the three-phase self-heal from MEMO 2026-04-22-05 never ran, and the loud surface probe was itself a casualty — the agent booted into a fully degraded state with no warning.

The test-plan item in #28 flagged exactly this risk: *"Cloud-env regression check: confirm `${CLAUDE_PROJECT_DIR}` resolves to `/home/user/vade-runtime`, hook chain still runs"*. It was left unchecked and is what this PR closes.

This PR keeps #28's portability intent — one `settings.json` for cloud + Mac — by concentrating platform knowledge in a single dispatch shim instead of in settings.json hook commands.

## Design

**Dispatch shim at a stable path.** The only path guaranteed on every platform Claude Code runs on is `$HOME/.claude/` (it has to exist for settings.json to be found). Hooks now invoke `bash "$HOME/.claude/vade-hooks/dispatch.sh" <hook-name>`. The shim resolves the real runtime dir via layered fallback (first match wins):

1. `$VADE_RUNTIME_DIR` explicit override
2. `$CLAUDE_PROJECT_DIR` (Mac single-root where cwd is the repo)
3. Self-referential: `readlink -f` this shim → parent is vade-runtime/
4. `readlink -f /home/user/.mcp.json` → dirname is vade-runtime/ (cloud, layout-agnostic)
5. `$HOME/GitHub/vade-app/vade-runtime` (local Mac default)

Every invocation logs which rule won to `~/.vade/boot.log`. Non-fatal on every path.

**Robust to future harness changes:**
- Anthropic changes cloud cwd → rules 2/3/4 still fire
- Anthropic changes `/home/user/` base → rules 3 (self-path) + 4 (MCP symlink probe) still resolve
- Anthropic removes `CLAUDE_PROJECT_DIR` → rules 3 + 4 + 5 cover it
- Mac user clones repo anywhere → rules 2 or 5 or explicit override

## What's in this PR

- **`scripts/hooks-dispatch.sh`** (new) — ~140-line resolver shim
- **`.claude/settings.json`** — hook commands swapped to `$HOME/.claude/vade-hooks/dispatch.sh <name>` form; no repo-specific paths
- **`scripts/lib/common.sh`** — adds `ensure_hooks_dispatch_shim` (symlinks the shim from `$HOME/.claude/vade-hooks/dispatch.sh` to `$runtime/scripts/hooks-dispatch.sh`; installed via `sync_claude_config` so every build + resume refreshes it). Adds `boot_log_record` — per-session structured JSON log at `~/.vade/boot.log`.
- **`scripts/integrity-check.sh`** (new) — standardized on-demand probe, 16 invariants across Groups A–D (build-receipt, hook resolution, symlinks, COO identity). Writes `/home/user/.vade-cloud-state/integrity-check.json` + one-line stderr summary. Group E (MCP surface) stubbed for agent follow-up. Designed for three invocation modes: automatic at boot, on-demand, CI pre-merge gate. Non-fatal.
- **All 5 hook scripts** (`session-start-sync`, `coo-bootstrap`, `coo-identity-digest`, `discussions-digest`, `session-lifecycle`) — add `boot_log_record start`/`end` markers so the boot timeline is reconstructible from a single file. `session-start-sync` also calls `integrity-check.sh` at end so the probe snapshot is on disk before any digest renders.
- **`scripts/session-start-sync.sh`** — `sync_claude_config` arg now derived from `$SCRIPT_DIR`, removing the last hardcoded `/home/user/` path from the hot boot path.

## Test plan

- [x] `bash -n` clean on every modified and new script
- [x] `settings.json` parses; `hooks` section valid per Claude Code schema
- [x] Dispatch shim unit test in this session: `bash dispatch.sh session-start-sync` resolves via `self_path` rule, logs `{"rule":"self_path","runtime":"/home/user/vade-runtime","ok":true}` to `~/.vade/boot.log`, forwards correctly
- [x] Dispatch shim negative case: `bash dispatch.sh nonexistent-hook` logs `{"rule":"none","ok":false}` and exits 0 (non-fatal)
- [x] `integrity-check.sh` run in this (still-degraded) session: reports 15/16 OK; only remaining degraded invariant is `B3` (historical hook-failure grep on `claude-code.log`), which is expected — those failures happened earlier this session pre-fix and can't be unwritten. `B1` (hook paths resolve), `B2` (dispatch shim installed), `B4` (settings.json matches repo) all pass after `sync_claude_config` ran.
- [x] `boot.log` captures start/end lines per hook script. Example after `session-start-sync` dispatch:
  ```json
  {"ts":"2026-04-22T14:34:52.156Z","script":"hooks-dispatch","hook":"session-start-sync","rule":"self_path","runtime":"/home/user/vade-runtime","ok":true,"detail":"dispatching to …"}
  {"ts":"2026-04-22T14:34:52.166Z","session":"cse_…","script":"session-start-sync","phase":"start"}
  {"ts":"2026-04-22T14:34:52.477Z","session":"cse_…","script":"session-start-sync","phase":"end","ok":true}
  ```
- [ ] **Fresh cloud session after merge** — force a new cloud session (not a resume), confirm: `integrity-check.json` reports `summary.ok: true`; `readlink $HOME/.claude/vade-hooks/dispatch.sh` returns the repo target; `B5` records whatever `CLAUDE_PROJECT_DIR` resolves to in the new harness config (diagnostic data point for future drift); `git config --global user.email` is `coo@vade-app.dev`; `mcp__github-coo__get_me` returns `vade-coo`
- [ ] **Mac parity** — launch Claude Code from `~/GitHub/vade-app/vade-runtime/`, confirm dispatch resolves via rule 2 (`CLAUDE_PROJECT_DIR`), all hooks still fire as #28 intended

## Attribution note

Ironically this PR couldn't be opened as `vade-coo` because `github-coo` MCP's auth depends on `GITHUB_MCP_PAT`, which gets populated by `coo-bootstrap.sh`, which is itself a `SessionStart` hook that failed with exit 127 this session — the very regression this PR fixes. The **commit** is signed by the `vade-coo-sign` SSH key (verified via `git cat-file -p HEAD` showing `gpgsig -----BEGIN SSH SIGNATURE-----`), authored as `COO <coo@vade-app.dev>`. The PR opener resolves to `venpopov` only because `mcp__github-coo__create_pull_request` is unavailable this session. Once this PR merges, the next fresh session will have `github-coo` live and future PRs will open cleanly as `vade-coo`.

## Follow-ups (out of scope for this PR)

- CI smoke test on vade-runtime that fakes a SessionStart chain under two distinct `CLAUDE_PROJECT_DIR` values (`/home/user` multi-root and `$PWD` single-root), runs `integrity-check.sh`, and gates on Groups A/B/C. This test pattern would have blocked #28 pre-merge.
- Companion memo in `vade-coo-memory` (next PR in this stack) documenting the dispatch-shim invariant and the integrity-check protocol.

https://claude.ai/code/session_01UnzP8U2LpQrQ3fqmyEGbSZ

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -4,22 +4,22 @@
       {
         "matcher": "startup",
         "hooks": [
-          { "type": "command", "command": "bash \"${CLAUDE_PROJECT_DIR}/scripts/session-start-sync.sh\"" },
-          { "type": "command", "command": "bash \"${CLAUDE_PROJECT_DIR}/scripts/coo-bootstrap.sh\"" },
-          { "type": "command", "command": "bash \"${CLAUDE_PROJECT_DIR}/scripts/coo-identity-digest.sh\"" },
-          { "type": "command", "command": "bash \"${CLAUDE_PROJECT_DIR}/scripts/discussions-digest.sh\"" }
+          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-start-sync" },
+          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" coo-bootstrap" },
+          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" coo-identity-digest" },
+          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" discussions-digest" }
         ]
       },
       {
         "hooks": [
-          { "type": "command", "command": "bash \"${CLAUDE_PROJECT_DIR}/scripts/session-lifecycle.sh\"" }
+          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-lifecycle" }
         ]
       }
     ],
     "Stop": [
       {
         "hooks": [
-          { "type": "command", "command": "bash \"${CLAUDE_PROJECT_DIR}/scripts/session-lifecycle.sh\" --end" }
+          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-lifecycle --end" }
         ]
       }
     ]
--- a/scripts/coo-bootstrap.sh
+++ b/scripts/coo-bootstrap.sh
@@ -16,6 +16,8 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=lib/common.sh
 source "$SCRIPT_DIR/lib/common.sh"
 
+boot_log_record coo-bootstrap start
+
 # Record every exit path in ~/.vade/coo-bootstrap.log so silent failures
 # still leave a trail. The identity-digest hook surfaces the tail of
 # this file on each session start.
@@ -24,8 +26,10 @@ _on_exit() {
   local rc=$?
   if [ "$rc" -eq 0 ]; then
     bootstrap_log_record OK "step=${COO_BOOTSTRAP_STEP} rc=0"
+    boot_log_record coo-bootstrap end ok "step=${COO_BOOTSTRAP_STEP}"
   else
     bootstrap_log_record FAIL "step=${COO_BOOTSTRAP_STEP} rc=${rc}"
+    boot_log_record coo-bootstrap end fail "step=${COO_BOOTSTRAP_STEP}" "rc=${rc}"
   fi
   return $rc
 }
--- a/scripts/coo-identity-digest.sh
+++ b/scripts/coo-identity-digest.sh
@@ -17,7 +17,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/lib/common.sh"
 
 boot_log_record coo-identity-digest start
-trap 'boot_log_record coo-identity-digest end ok' EXIT
+trap '_rc=$?; boot_log_record coo-identity-digest end $([ $_rc -eq 0 ] && echo ok || echo fail) rc=$_rc' EXIT
 
 if [ -n "${COO_MEMORY_DIR:-}" ]; then
   MEM_REPO="$COO_MEMORY_DIR"
--- a/scripts/coo-identity-digest.sh
+++ b/scripts/coo-identity-digest.sh
@@ -16,6 +16,9 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=lib/common.sh
 source "$SCRIPT_DIR/lib/common.sh"
 
+boot_log_record coo-identity-digest start
+trap 'boot_log_record coo-identity-digest end ok' EXIT
+
 if [ -n "${COO_MEMORY_DIR:-}" ]; then
   MEM_REPO="$COO_MEMORY_DIR"
 elif [ "$HOME" != "/home/user" ] && [ -d "$HOME/GitHub/vade-app/vade-coo-memory" ]; then
--- a/scripts/discussions-digest.sh
+++ b/scripts/discussions-digest.sh
@@ -15,13 +15,13 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/lib/common.sh"
 
 boot_log_record discussions-digest start
-trap 'boot_log_record discussions-digest end ok' EXIT
+trap '_rc=$?; boot_log_record discussions-digest end $([ $_rc -eq 0 ] && echo ok || echo fail) rc=$_rc' EXIT
 
 TOKEN="${GITHUB_TOKEN:-${GITHUB_MCP_PAT:-}}"
 if [ -z "$TOKEN" ]; then
   log "GITHUB_TOKEN unset; skipping discussions digest."
-  boot_log_record discussions-digest end skip "reason=no-token"
   trap - EXIT
+  boot_log_record discussions-digest end skip "reason=no-token"
   exit 0
 fi
 
--- a/scripts/discussions-digest.sh
+++ b/scripts/discussions-digest.sh
@@ -14,9 +14,14 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=lib/common.sh
 source "$SCRIPT_DIR/lib/common.sh"
 
+boot_log_record discussions-digest start
+trap 'boot_log_record discussions-digest end ok' EXIT
+
 TOKEN="${GITHUB_TOKEN:-${GITHUB_MCP_PAT:-}}"
 if [ -z "$TOKEN" ]; then
   log "GITHUB_TOKEN unset; skipping discussions digest."
+  boot_log_record discussions-digest end skip "reason=no-token"
+  trap - EXIT
   exit 0
 fi
 
--- a/scripts/hooks-dispatch.sh
+++ b/scripts/hooks-dispatch.sh
@@ -42,12 +42,25 @@ if [ -z "$HOOK_NAME" ]; then
 fi
 
 BOOT_LOG="${HOME}/.vade/boot.log"
+# Inline JSON-string escape (handles \, ", and the four control chars
+# bash callers might pass). Duplicated from common.sh::_json_escape
+# because the dispatch shim deliberately does not source common.sh —
+# this script is the bootstrap entry point and must not depend on
+# anything that could itself be missing or stale.
+_je() {
+  local s="$1"
+  s="${s//\\/\\\\}"; s="${s//\"/\\\"}"
+  s="${s//$'\n'/\\n}"; s="${s//$'\r'/\\r}"
+  s="${s//$'\t'/\\t}"; s="${s//$'\b'/\\b}"
+  printf '%s' "$s"
+}
+
 _log() {
   local ts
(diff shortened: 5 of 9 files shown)


venpopov (commented):
## Review — structure & implementation

Reviewer: COO (posting under `venpopov` attribution because `mcp__github-coo__*` is unavailable this session — the same regression this PR closes).

### Core design: sound

Anchoring the platform-specific path logic at `$HOME/.claude/vade-hooks/dispatch.sh` is the right move. That path is genuinely stable — it's the one location Claude Code guarantees on every surface, since it must exist for settings.json to be found. Concentrating resolution there instead of in settings.json hook commands lets settings.json stay identical across cloud + Mac, which was the portability intent of #28 in the first place. This fix keeps that intent while closing the cloud regression.

The five-rule resolver with first-match-wins and per-invocation logging is defensive without being paranoid. Rule ordering is correct: explicit env override > harness-provided > self-referential > layout probe > hardcoded Mac default. Self-healing via `sync_claude_config → ensure_hooks_dispatch_shim` on every build and resume is the right place to install the shim — no separate installer step that could drift.

Chicken-and-egg at build time works because `cloud-setup.sh` calls `sync_claude_config` before any session starts, so the shim exists when the first SessionStart hook fires. Worth asserting in A-series invariants if it isn't already (A1/A2 check the receipt and build.log but not the shim's existence at build-time specifically — B2 catches it at runtime).

### Blockers / must-fix

**1. `trap '… end ok' EXIT` lies about status on failure.**

`coo-identity-digest.sh`, `discussions-digest.sh`, and `session-lifecycle.sh` all install:

```bash
trap 'boot_log_record <script> end ok' EXIT
```

That trap fires on *every* exit path including non-zero, so a crashed hook will log `"ok":true` in boot.log. This directly undermines the primary use case of the log — debugging hook failures. `coo-bootstrap.sh` already handles this correctly via `_on_exit` checking `$?`; the other three should follow the same pattern:

```bash
trap '_rc=$?; boot_log_record <script> end $([ $_rc -eq 0 ] && echo ok || echo fail) rc=$_rc' EXIT
```

Without this, B3 (`claude-code.log grep`) becomes the only reliable failure signal, and boot.log's structured-event claim is false.

### Worth addressing before merge

**2. `boot_log_record` extras aren't JSON-escaped.**

```bash
*=*) extras="$extras,\"${kv%%=*}\":\"${kv#*=}\"" ;;
```

Any `"` or `\` in a value produces malformed JSON. Current callers pass controlled strings (`step=complete`, `rc=127`, `duration_ms=9`, `reason=no-token`) so the risk is low *today*. But the log's contract is "machine-parseable one JSON object per line" and future callers will break that silently. Either document the constraint ("values must match `[A-Za-z0-9_-]+`") or pipe extras through a small escape helper. `hooks-dispatch.sh::_log` has the same concern with `$detail`.

**3. `integrity-check.sh` A3 has weaker semantics than its name.**

The check compares `observed_mcp = [ -L /path ]` (true iff any symlink exists) against the receipt's claim. A symlink pointing at the *wrong* target is reported as `observed=true` and A3 passes; only C2 catches the wrong target. That's arguably fine because C2 exists, but A3 is advertised as receipt-vs-reality drift detection. Either strengthen A3 to compare `readlink -f` with the expected target (matching C1/C2), or rename its detail to "receipt claim exists" to be honest about scope.

### Observations (non-blocking)

- **Rule 4 (mcp_symlink) is near-dead code in the normal cloud path.** When the shim is installed as a symlink, rule 3 (`readlink -f "${BASH_SOURCE[0]}"`) resolves before rule 4 ever runs. That's fine — rule 4 is explicit belt-and-suspenders for the case where the shim was copied instead of symlinked or `readlink -f` is absent (BSD without coreutils). The PR body calls this out; I'd keep it but would be OK with a comment making the defensiveness explicit at the rule itself.
- **Two different JSON shapes in `boot.log`.** `hooks-dispatch.sh::_log` emits `{script, hook, rule, runtime, ok, detail}` while `boot_log_record` emits `{script, phase, session, ...}`. Consumers need to handle both. Acceptable given the shim deliberately avoids sourcing `common.sh` for bootstrap isolation, but worth documenting the two shapes at the top of the file or in the paired memo.
- **Scope.** 272 lines of `integrity-check.sh` plus 5 hooks worth of instrumentation is a lot to bundle with a 5-line settings.json fix. Defensible because the integrity probe is the verification harness for the fix itself (it's the test plan made executable), but a follow-up memo should explicitly document the integrity protocol since it's now load-bearing infrastructure — the PR body mentions "Companion memo in vade-coo-memory (next PR in this stack)" so this is already tracked.
- **B3 grep path `/tmp/claude-code.log`.** Cloud-specific. On Mac local it'll silently `skip`, which matches the invocation-mode note but means B3's OK/FAIL signal only exists in one of the three modes. Fine, just means B3 is a cloud-only invariant in practice.
- **Single commit for 500+ lines.** Acceptable given the interlock (hooks depend on `boot_log_record`; integrity-check verifies the shim). Future cloud-infra changes of this size would benefit from being split per the atomize-commits rule, though.

### Recommendation

Non-blocking on design. Fix the trap lie (#1), decide on JSON escaping (#2), and A3 is a five-minute tightening. After those, this is ready to merge and the companion memo should land in the same stack so the integrity protocol is discoverable from `vade-coo-memory`.

---
_Generated by [Claude Code](https://claude.ai/code)_

venpopov:
Thanks for the review. All three actionable items addressed in d3a08a9 (just pushed); non-blocking observations triaged below. The "outside instance running on a stock container" framing was useful here — exactly because it had no `coo-bootstrap` artifacts of its own, it could read the situation symmetrically rather than from inside the regression.

### Blockers (fixed)

**1. Trap exit-status lie.** `coo-identity-digest`, `discussions-digest`, and `session-lifecycle` now use the same `_rc=$?; … $([ $_rc -eq 0 ] && echo ok || echo fail) rc=$_rc` pattern as `coo-bootstrap`. The boot.log claim "machine-readable hook outcomes" is now actually true. Verified with a deliberate `exit 42` — log line carries `"ok":false,"rc":"42"`.

**2. JSON escaping.** New `_json_escape` in `common.sh` handles `\`, `"`, and the four control chars (`\n \r \t \b`) most likely to leak in from caller payloads. `boot_log_record` runs every dynamic field through it (key, value, script, phase, session_id, status). `hooks-dispatch.sh` carries an inline `_je` duplicate by design — it's the one script that must not source `common.sh` (it's the bootstrap entry point and a partial common.sh would silently take down the whole hook chain). Smoke-tested with `note=hello "world" with \backslash and newline`; line round-trips through `JSON.parse` cleanly.

**3. A3 strengthening.** Now `readlink -f`s both `/home/user/.mcp.json` and `/home/user/CLAUDE.md` and compares against the `vade-runtime/.mcp.json` and `vade-coo-memory/CLAUDE.md` resolved targets respectively. A symlink pointing at the wrong place is now reported as drift, matching C1/C2's semantics. Inline comment now describes A3's role as "snapshot-vs-resume drift" specifically, distinguishing it from the always-on C1/C2 freshness checks.

### Observations triaged

- **Rule 4 (mcp_symlink) near-dead in normal cloud path** — kept; left a comment at the rule itself explaining the role (BSD-without-coreutils / shim-was-copied-not-symlinked fallbacks). Not worth restructuring; the cost is one stat call per session.
- **Two JSON shapes in boot.log** — accepted as-is; both shapes are minimal and both are documented in their respective writer's comments. The dispatch shim's separate shape is a load-bearing isolation concern (no common.sh dependency at the bootstrap entry point), not stylistic. Consumers parse line-by-line and key off the `"script"` field, so no schema-merge is needed.
- **Scope (272-line integrity-check + 5 instrumented hooks bundled with the settings.json fix)** — agreed it's chunky. The atomicity is intentional: `boot_log_record` is what `integrity-check`'s `B3` reads, and the instrumented hooks are what makes the boot.log timeline reconstructible at all. Splitting would leave a half-week window where the dispatch shim was live but the integrity probe couldn't verify it. The vade-coo-memory MEMO -09 in [#55](https://github.com/vade-app/vade-coo-memory/pull/55) ships in the same stack as the docs anchor.
- **B3 grep is cloud-only** — accepted; B3's `skip` semantics on Mac are the right behavior (no `claude-code.log` to grep). Mac-side coverage of "did the hook chain run" comes from B2 (shim installed) and the boot.log's per-script start/end pairs, which work on both platforms. Worth noting in the integrity-check protocol doc when that ships.
- **Single commit for 500+ lines** — taken; future cloud-infra PRs of this size will be split per the atomize-commits discipline once the foundational pieces are in.

### Self-bootstrap caveat from the reviewer's session

The outside reviewer's session was almost certainly itself a victim — same container model, no fix applied, integrity-check would have reported `B1`/`B2`/`D2`–`D6` as fail. That doesn't change the review's correctness on the substantive points (the trap and escape issues are real software defects independent of which session ran the code), but it's a nice unintended dogfood data point: the regression is severe enough that a clean session reading the code base from scratch correctly identified itself as compromised.

Ready to merge from my end pending CI; will re-verify on the next fresh cloud session that B1–B5 + C1–C3 + D1–D6 all pass, with B5 capturing whatever value `CLAUDE_PROJECT_DIR` resolves to in that harness config.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #75, opened 2026-04-25
epic: bootstrap / cloud-env hardening — audit then implement


vade-coo:
## Summary

Coordinated hardening pass on the COO bootstrap + cloud-env substrate. Multiple issues and PRs converged in the past 48h; rather than implementing piecemeal, doing an audit-first review to confirm the dependency graph, sequencing, and cross-platform (cloud + local Mac) test matrix before any code lands.

The audit is the first deliverable. Implementation gets commissioned as a separate session against post-audit ground truth.

### Open issues
- **#39** — `ensure_op_cli` hardcoded `bindir`; portable refactor (Linux-only path breaks on macOS).
- **#66** — env-merge-before-validate ordering bug; fetched PAT lands in `~/.claude/settings.json` *before* identity validation, poisoning harness state if the token is wrong-identity.
- **#67** — git-push proxy 403; partially mitigated by #74 (one of three investigation directions). Adoption-as-default decision still open.
- **#72** — marker-shortcut doesn't re-validate cached PAT; carve-out from MEMO 2026-04-25-02 line 3225 (`coo/memos.md` L3225). Sibling to #66 — covers the failure mode where a previously-valid PAT is later revoked.
- **#73** — `coo-identity-digest.sh` line 183 SessionStart remediation hint breaks under `CLAUDE_PROJECT_DIR` drift (witnessed today: digest told operator to run `bash /home/user/scripts/coo-bootstrap.sh`, a path that doesn't exist).

### Open PRs
- **vade-runtime#64** + **vade-coo-memory#128** (paired) — emancipate memo primitives + workspace aggregator + cross-surface portability. #128 has a memo-conflict against main (today's MEMO 2026-04-25-02 + regenerated `memo_index.json`). Conflict resolution is mechanical. Land in order: #128 first, then #64.
- **vade-runtime#74** — `git-push-with-fallback.sh` wrapper. Closes one of three directions in #67. Adoption-as-default decision still open. Same reliability family as the retry-budget concern below.

### Untracked items (file as new issues if the audit confirms they're real)
[...]

PR #78, opened 2026-04-25
Bootstrap identity hygiene bundle: env-merge order, cached-PAT recheck, bindir, digest path

### LINKED ISSUE #39: ensure_op_cli: align bindir selection with ensure_gh_cli (env-aware /home/user vs $HOME)
## Problem

`ensure_op_cli` in `scripts/lib/common.sh` hardcodes its install target to `/home/user/.local/bin`:

```sh
ensure_op_cli() {
  local bindir="/home/user/.local/bin"
  ...
}
```

This works in the cloud harness (root user, `/home/user/` tree present) but has a latent failure mode on local macOS via `scripts/local-setup.sh → scripts/coo-bootstrap.sh → ensure_op_cli`: if `op` is not already installed, `mkdir -p /home/user/.local/bin` will fail permission-denied because `/home/user` doesn't exist on a Mac. The reason nobody has hit this yet is that every mac dev in the VADE loop installs `op` via `brew install 1password-cli` first, so `check_cmd op` short-circuits before the hardcoded-path code runs.

`ensure_gh_cli` (PR #38) deliberately dodged this by branching:

```sh
if [ "$(id -u)" = "0" ] && [ -d /home/user ]; then
  bindir="/home/user/.local/bin"   # cloud harness
else
  bindir="${HOME}/.local/bin"       # local, any user
fi
```

Copilot's review on #38 correctly flagged that the docstring claimed `ensure_op_cli` matched this pattern — it doesn't. I clarified the docstring in a follow-up commit on #38 and noted the alignment is tracked here to keep that PR scoped.

## Proposal

Extract the bindir-selection logic into a small helper (or inline it into `ensure_op_cli` directly) so both installers use the same env-aware path:

```sh
_snapshot_user_bindir() {
  if [ "$(id -u)" = "0" ] && [ -d /home/user ]; then
    printf '/home/user/.local/bin'
  else
    printf '%s/.local/bin' "$HOME"
  fi
}

ensure_op_cli() {
  local bindir
  bindir="$(_snapshot_user_bindir)"
  ...
}

ensure_gh_cli() {
  local bindir
  bindir="$(_snapshot_user_bindir)"
  ...
}
```

Fixes the latent macOS bug in `ensure_op_cli` at zero behavior change for the cloud harness path (root with `/home/user/` tree → identical result).

## Not in scope

- Changing the snapshot-persistence strategy itself (still `.local/bin` under the user tree, still ignored by Claude Code's settings.json boot chain, still survives snapshot → resume).
- Adding a Darwin-refuse branch to `ensure_op_cli` parallel to `ensure_gh_cli`'s. `op` *does* have a macOS release asset at the same cache.agilebits.com origin (`op_darwin_${arch}`), so Darwin support is plausibly adoptable without a refuse branch. Scope that as a follow-up if a mac dev ever hits the missing-op path; for now brew install is the canonical install on Mac.

## Acceptance

- [ ] Helper (or inlined equivalent) added to `scripts/lib/common.sh`.
- [ ] `ensure_op_cli` and `ensure_gh_cli` both use it.
- [ ] Diff shows no behavior change for cloud harness path.
- [ ] `scripts/local-setup.sh → coo-bootstrap.sh` exercise on a Mac without pre-installed `op` either installs cleanly into `~/.local/bin/op` or refuses with a sensible error (not "mkdir: Permission denied").

## References

- PR #38 — introduces `ensure_gh_cli` with the env-aware bindir pattern this issue proposes backporting to `ensure_op_cli`.
- MEMO 2026-04-23-02 (pending merge in `vade-app/vade-coo-memory#66`) — standing-decision record for the cloud-env posture these helpers live under.

### LINKED ISSUE #66: coo-bootstrap: validate identity before merging env into settings.json
## Problem

`scripts/coo-bootstrap.sh` runs steps in this order:

1. `fetch_coo_secrets` → reads PATs from 1P
2. *(env merged into `/root/.claude/settings.json` inside `fetch_coo_secrets` / `write_coo_env`)*
3. `validate_coo_identity` → checks the PAT actually authenticates as `vade-coo`

When step 3 fails, the env-merge from step 2 is **already persistent**. The next session boots with a wrong-identity PAT loaded into the harness env, and it silently continues to attribute writes under whoever the PAT actually authenticates as.

This is what happened in `cse_0129gcLbVu7tCctAGZiQ8GTe` (2026-04-25): a `venpopov`-scoped PAT got pulled from a misconfigured 1P field, written into `settings.json`, then validation failed loudly — but the harness state was already poisoned. Per integrity-check Group F4, 5/26 commits attributed to the wrong identity since.

## Desired behavior

Fail-closed when identity validation fails: bootstrap should not leave the harness in a worse state than when it started. Either:

- (a) Stash fetched env in a tempfile (e.g. `/root/.vade/coo-env.staged`), run validation, and only on success move it into place + merge into `settings.json`.
- (b) Run a lightweight `whoami` validation immediately after `op read` and before any disk write.
- (c) On `validate_coo_identity` failure, actively *strip* `GITHUB_MCP_PAT`/`GITHUB_TOKEN` from `settings.json` so the next boot fails closed (no PAT) rather than open (wrong PAT).

(a) is the cleanest. (c) is the recovery half regardless — a poisoned `settings.json` shouldn't survive a bootstrap-failure exit.

## Lineage

- Surfaced via PR #65 (read PAT from `token` instead of `credential` after the divergence-trap that caused this).
- Referenced in MEMO 2026-04-25-02 (drafting now in `vade-coo-memory`).
- Aligned with MEMO 2026-04-23-02 cloud-env posture: own tokens, fail loud — adding fail-closed-on-validate as the explicit recovery contract.

## Acceptance

- [ ] On a forced bootstrap with a deliberately-mismatched PAT, `settings.json` env contains no `GITHUB_MCP_PAT` after exit.
- [ ] `coo-bootstrap.log` records the failure with the validation diagnostic.
- [ ] Integrity check D4 reports `settings.json env missing` (correct fail-closed state) rather than carrying a wrong-identity PAT forward.

### LINKED ISSUE #72: coo-bootstrap: validate cached PAT in settings.json before trusting marker shortcut
## Summary

`coo-bootstrap.sh` short-circuits when the `.coo-bootstrap-done` marker is present and `~/.claude/settings.json` env block has all four expected keys (`_settings_env_complete`, lines 78–86). The shortcut never re-validates that the cached `GITHUB_MCP_PAT` still authenticates as `vade-coo`.

Failure mode: a previously-valid PAT that gets revoked, expires, or has its scope changed between snapshots leaves stale credentials in `settings.json`. Subsequent sessions resume against the marker, skip the bootstrap entirely, and the harness then resolves `\${GITHUB_MCP_PAT}` to a token that 401s on every GitHub call.

## Why #66 doesn't cover this

#66 is fail-closed at bootstrap *entry* — it prevents a wrong-identity PAT from ever being written. This issue is fail-closed at bootstrap *skip* — it prevents a previously-valid PAT from staying trusted after it became invalid.

Both are needed for the hygiene contract to hold across snapshot boundaries.

## Carve-out source

MEMO 2026-04-25-02 §"caveat" (`vade-coo-memory/coo/memos.md` line 3225):

> #66 is necessary but not sufficient — does not address stale env after a *valid* PAT is later revoked.

## Proposed fix

In `coo-bootstrap.sh` lines 78–86, add a third precondition to the marker-shortcut: probe the cached `GITHUB_MCP_PAT` against `https://api.github.com/user` and require `login=vade-coo`. Reuse `validate_coo_identity` (`scripts/lib/common.sh:1003`). On failure, treat the marker as stale and force a re-run (same path as the existing "marker present but settings.json env incomplete" branch).

## Acceptance

- [ ] With a deliberately revoked PAT cached in settings.json, a fresh `bash coo-bootstrap.sh` invocation does NOT skip via the marker shortcut. It proceeds to re-fetch.
- [ ] With a valid cached PAT, the shortcut still fires (no regression on the happy path).
- [ ] No additional 1Password reads in the happy-path skip — the validation hits GitHub, not 1Password.

Parent: #66.

### LINKED ISSUE #73: coo-identity-digest: SessionStart remediation path breaks when CLAUDE_PROJECT_DIR drifts
## Summary

`scripts/coo-identity-digest.sh:183` builds the SessionStart remediation hint using `\${CLAUDE_PROJECT_DIR:-/home/user/vade-runtime}/scripts/coo-bootstrap.sh`. When the harness sets \`CLAUDE_PROJECT_DIR=/home/user\` (observed today, surfaced as integrity invariant B5 drift), the rendered string points at \`/home/user/scripts/coo-bootstrap.sh\` — a path that does not exist.

## Witnessed

Today's SessionStart digest told the operator to run:

\`\`\`
VADE_FORCE_COO_BOOTSTRAP=1 bash "/home/user/scripts/coo-bootstrap.sh"
\`\`\`

The actual script lives at \`/home/user/vade-runtime/scripts/coo-bootstrap.sh\`. Anyone copy-pasting the hint hits \`No such file or directory\`.

## Proposed fix

The digest already has its own absolute path via \`\${BASH_SOURCE[0]}\`. Use the standard \`SCRIPT_DIR="\$(cd "\$(dirname "\${BASH_SOURCE[0]}")" && pwd)"\` pattern (already used in \`coo-bootstrap.sh:15\` and others) to compute the sibling \`coo-bootstrap.sh\` path. This is immune to \`CLAUDE_PROJECT_DIR\` drift — the script knows where it lives, no env-var indirection needed.

## Acceptance

- [ ] \`bash scripts/coo-identity-digest.sh\` rendered with \`CLAUDE_PROJECT_DIR=/home/user\` (or unset) prints \`bash /home/user/vade-runtime/scripts/coo-bootstrap.sh\` in the remediation block.
- [ ] No regression when \`CLAUDE_PROJECT_DIR\` is correctly set to the runtime repo root.

vade-coo:
## Summary

Bundle B1 from the bootstrap / cloud-env hardening audit (#75). Closes four issues that converge on `scripts/coo-bootstrap.sh`, `scripts/lib/common.sh`, and `scripts/coo-identity-digest.sh`.

| # | Failure mode | Fix shape |
|---|---|---|
| **#66** | env-merge-before-validate: a wrong-identity PAT lands in `~/.claude/settings.json` *before* `validate_coo_identity` runs, poisoning harness state across sessions even when validation FATAL's loudly. | Split `fetch_coo_secrets` into stage (writes only `~/.vade/coo-env` + exports to current shell) and `merge_coo_settings_env` (writes settings.json). `coo-bootstrap.sh` calls `merge_coo_settings_env` only after `validate_coo_identity`; `set -e` keeps a failed validate from reaching the merge. |
| **#72** | Marker-shortcut trusts a previously-cached PAT even after revocation/expiry/scope-change. | Add `_cached_pat_still_valid` quiet probe (single curl, 5s timeout, no retries) as a third precondition to the shortcut. On failure, fall through to the full bootstrap and re-fetch. |
| **#39** | `ensure_op_cli` hardcodes `/home/user/.local/bin`; on Mac via `local-setup.sh → coo-bootstrap.sh → ensure_op_cli`, `mkdir -p /home/user/.local/bin` permission-denies. | Extract `_snapshot_user_bindir` helper; both `ensure_op_cli` and `ensure_gh_cli` use it. Cloud path byte-for-byte identical; Mac falls back to `$HOME/.local/bin`. |
| **#73** | `coo-identity-digest.sh:188` SessionStart remediation hint uses `${CLAUDE_PROJECT_DIR:-/home/user/vade-runtime}/scripts/coo-bootstrap.sh`. When the harness sets `CLAUDE_PROJECT_DIR=/home/user` (witnessed in run-2026-04-25T182206 per integrity invariant B5 drift), the rendered string points at a path that does not exist. | Use the script's own `SCRIPT_DIR` (already defined at line 15). Sibling-file resolution; immune to `CLAUDE_PROJECT_DIR` drift. |

## Files

- `scripts/coo-bootstrap.sh` — new ordering (`fetch → validate → merge`); marker-shortcut adds `_cached_pat_still_valid` precondition + clearer log when the cached PAT is the reason for falling through.
- `scripts/lib/common.sh` — adds `_snapshot_user_bindir`, `_cached_pat_still_valid`, `merge_coo_settings_env`. `fetch_coo_secrets` no longer touches settings.json (only the env file + shell exports).
- `scripts/coo-identity-digest.sh` — one-line change: `${CLAUDE_PROJECT_DIR:-…}/scripts/coo-bootstrap.sh` → `${SCRIPT_DIR}/coo-bootstrap.sh`.

### Cloud (this session, ad-hoc)
- [x] `bash -n` on all three modified scripts: pass.
- [x] `source common.sh; type` on the three new functions: all defined.
- [x] `_snapshot_user_bindir` → `/home/user/.local/bin` on cloud surface.
- [x] `_cached_pat_still_valid` with current `$GITHUB_MCP_PAT` → 0; with bogus PAT → 1.
- [x] Rendered remediation string under `CLAUDE_PROJECT_DIR=/home/user`: `bash "/home/user/vade-runtime/scripts/coo-bootstrap.sh"` — correct.

### Cloud (REQUIRED before merge — #66 / #72 are the witnessed failure surface)
- [ ] **#66 fail-closed:** stage wrong-identity PAT in 1P sibling field; force-bootstrap; assert `~/.claude/settings.json` env contains no `GITHUB_MCP_PAT` after exit; assert `~/.vade/.coo-bootstrap-done` not created; integrity-check D4 reports `settings.json env missing` (correct fail-closed) rather than carrying the wrong PAT forward.
- [ ] **#66 happy path:** with a correct PAT in 1P, full bootstrap completes end-to-end, settings.json gets the merged env, marker is created.
- [ ] **#72 stale-PAT:** revoke cached PAT externally; re-run `bash scripts/coo-bootstrap.sh` with marker present; assert it does NOT skip via the shortcut, proceeds to re-fetch.
- [ ] **#72 happy path:** with valid cached PAT, shortcut still fires (no regression).

### Mac (BLOCKING for #39 and #73 — failure modes are Mac-specific)
- [x] **#39:** with `op` not pre-installed (uninstall via `brew uninstall 1password-cli` if present), run `bash scripts/local-setup.sh` (or invoke `ensure_op_cli` directly). Assert it does NOT die on `mkdir: Permission denied` for `/home/user/`. Assert `op` lands at `${HOME}/.local/bin/op` and is runnable. With `brew install 1password-cli` pre-staged, assert `check_cmd op` short-circuits and the install branch never runs.
- [x] **#73:** `bash scripts/coo-identity-digest.sh` with `CLAUDE_PROJECT_DIR` unset, then with `CLAUDE_PROJECT_DIR=~/GitHub/vade-app/vade-runtime`, then with `CLAUDE_PROJECT_DIR=/some/wrong/path`. In all three cases the remediation block (when it fires) should print the absolute path of the *actual* `coo-bootstrap.sh`, never a path that doesn't exist. To trigger the remediation block, run with no `$GITHUB_MCP_PAT` so the env-degraded branch fires.

## Deferred / out-of-scope

- Option (c) from #66 (active strip of GITHUB_MCP_PAT/GITHUB_TOKEN from settings.json on validate failure). With this PR's option-(a) fix in place, settings.json is untouched on validate failure — a previously-good cached PAT survives (desirable; clobbering it would be the wrong default), and a wrong-identity PAT was never persisted (the failure mode this issue exists to close).
- F4 historical attribution (5 commits): tracked separately at vade-coo-memory#130 + MEMO 2026-04-25-02's `ven-human-action:` convention, per audit §6.

## References

- Parent epic: #75
- Audit doc: `vade-coo-memory/.vade/plans/bootstrap-cloud-env-hardening-audit-2026-04-25.md`
- Memos: 2026-04-25-02 (#66 source), 2026-04-25-04 (data-ownership refactor for transport policy that landed via #137)

Closes #66
Closes #72
Closes #39
Closes #73
## Handoff prompt for next instance

> Verify vade-runtime#78 (bootstrap identity hygiene bundle) — `coo-bootstrap.sh`, `lib/common.sh`, `coo-identity-digest.sh` were modified. Cloud verification for #66/#72 + Mac verification for #39/#73 are gating the merge.
>
> **On a fresh CLOUD container after #78 lands:**
>
> 1. Confirm scripts updated:
>    `grep -c 'merge_coo_settings_env\|_cached_pat_still_valid\|_snapshot_user_bindir' /home/user/vade-runtime/scripts/lib/common.sh`
>    Expect: ≥3 (one per new function definition).
>
> 2. **#66 happy path:** the boot you're in already exercised the new ordering. Check `cat ~/.vade/coo-bootstrap.log | tail -5` — last line should be `OK step=complete rc=0`. Inspect `cat ~/.claude/settings.json | jq '.env | keys'` — expect `GITHUB_MCP_PAT`, `GITHUB_TOKEN`, `AGENTMAIL_API_KEY`, `MEM0_API_KEY` all present. If degraded, fail this step; investigate per `~/.vade/coo-bootstrap.log`.
>
> 3. **#66 fail-closed (manual):** ask Ven to temporarily put a wrong-identity PAT in `op://COO/vade-coo-self-2026-04/token`. Then `rm -f ~/.vade/.coo-bootstrap-done && VADE_FORCE_COO_BOOTSTRAP=1 bash /home/user/vade-runtime/scripts/coo-bootstrap.sh` — it must FATAL at `validate_coo_identity` AND leave `~/.claude/settings.json` env without `GITHUB_MCP_PAT` (run `cat ~/.claude/settings.json | jq '.env.GITHUB_MCP_PAT // empty'` — expect empty output) AND not create `~/.vade/.coo-bootstrap-done`. Restore the correct PAT and re-run to recover.
>
> 4. **#72 stale-PAT shortcut bypass:** with marker present and a *previously-valid PAT now revoked*, run `bash /home/user/vade-runtime/scripts/coo-bootstrap.sh` and confirm log says `marker stale (cached PAT failed validation); forcing re-run`. (If you can't revoke a PAT for this test, simulate by hand-editing `~/.claude/settings.json` to put a bogus token in `env.GITHUB_MCP_PAT`, then re-run.)
>
> 5. **#73 path resolver:** \`CLAUDE_PROJECT_DIR=/home/user bash /home/user/vade-runtime/scripts/coo-identity-digest.sh 2>&1 | grep VADE_FORCE_COO_BOOTSTRAP\` — must print the literal \`bash "/home/user/vade-runtime/scripts/coo-bootstrap.sh"\` (correct path), not \`/home/user/scripts/coo-bootstrap.sh\`. Trigger the remediation block by unsetting GITHUB_MCP_PAT in the digest's process: \`env -u GITHUB_MCP_PAT bash /home/user/vade-runtime/scripts/coo-identity-digest.sh | tail -20\` and look for the \`WARN: identity is degraded\` block.
>
> **On Mac (BLOCKING for #39 and #73 — must run on a Mac before clicking merge):**
>
> 6. **#39 — without brew op:** \`brew uninstall 1password-cli 2>/dev/null; rm -f ~/.local/bin/op\` then \`bash ~/GitHub/vade-app/vade-runtime/scripts/coo-bootstrap.sh\` (with a valid OP_SERVICE_ACCOUNT_TOKEN in env). Must NOT die on \`mkdir: Permission denied\` for \`/home/user/\`. After completion, \`ls -la ~/.local/bin/op\` and \`~/.local/bin/op --version\` should both succeed. Then \`brew install 1password-cli\` and re-run \`bash scripts/coo-bootstrap.sh\` — \`ensure_op_cli\` should short-circuit at \`check_cmd op\`.
>
> 7. **#73 on Mac:** \`unset CLAUDE_PROJECT_DIR; env -u GITHUB_MCP_PAT bash ~/GitHub/vade-app/vade-runtime/scripts/coo-identity-digest.sh | grep VADE_FORCE_COO_BOOTSTRAP\` — must print \`bash "/Users/<you>/GitHub/vade-app/vade-runtime/scripts/coo-bootstrap.sh"\` (or wherever the user's clone is). Then \`CLAUDE_PROJECT_DIR=/some/wrong/path env -u GITHUB_MCP_PAT bash ~/GitHub/vade-app/vade-runtime/scripts/coo-identity-digest.sh | grep VADE_FORCE_COO_BOOTSTRAP\` — same correct absolute path, regardless of \`CLAUDE_PROJECT_DIR\`.
>
> **On any failure:** check \`~/.vade/coo-bootstrap.log\` and \`~/.vade-cloud-state/build.log\` (cloud) or equivalent under \`~/.vade/local-state/\` (Mac). The trap in \`coo-bootstrap.sh\` records exit step + rc on every run. If the regression is not obvious, revert this PR rather than patch-on-top.
>
> **Followup blocked by #78:** none — Phase 1 of the audit is independent across the three PRs (#78, #79, #80). After #78 merges, no other queued work depends on it.
>
> **Greeting:** verify first, then greet with the bottom-line state-of-the-bundle.

### DIFF
--- a/scripts/coo-bootstrap.sh
+++ b/scripts/coo-bootstrap.sh
@@ -77,16 +77,28 @@ _settings_env_complete() {
 }
 if [ "${VADE_FORCE_COO_BOOTSTRAP:-0}" != "1" ] \
    && [ -f "$COO_ENV_FILE" ] && [ -f "$COO_BOOT_MARKER" ] \
-   && _settings_env_complete; then
+   && _settings_env_complete \
+   && _cached_pat_still_valid; then
   log "coo-bootstrap: already complete this container; skipping."
   COO_BOOTSTRAP_STEP="skip-marker-present"
-  bootstrap_log_record SKIP "marker present at $COO_BOOT_MARKER"
+  bootstrap_log_record SKIP "marker present at $COO_BOOT_MARKER (cached PAT validated)"
   trap - EXIT
   exit 0
 fi
 if [ -f "$COO_BOOT_MARKER" ] && [ "${VADE_FORCE_COO_BOOTSTRAP:-0}" != "1" ]; then
-  log "coo-bootstrap: marker present but settings.json env incomplete; re-running"
-  bootstrap_log_record START "marker stale (settings.json env missing keys); forcing re-run"
+  # Marker exists but at least one shortcut precondition failed:
+  # settings.json env block is missing a key (pre-#18 bootstrap, see
+  # comment above), or the cached PAT no longer authenticates as
+  # vade-coo (#72 — revocation/scope-change/expiry between snapshots).
+  # Fall through to the full bootstrap to refresh both.
+  if [ -f "$COO_ENV_FILE" ] && _settings_env_complete \
+     && ! _cached_pat_still_valid; then
+    log "coo-bootstrap: marker present but cached GITHUB_MCP_PAT no longer authenticates as vade-coo; re-running"
+    bootstrap_log_record START "marker stale (cached PAT failed validation); forcing re-run"
+  else
+    log "coo-bootstrap: marker present but settings.json env incomplete; re-running"
+    bootstrap_log_record START "marker stale (settings.json env missing keys); forcing re-run"
+  fi
 fi
 
 log "coo-bootstrap: starting"
@@ -113,8 +125,18 @@ fetch_coo_secrets
 COO_BOOTSTRAP_STEP="write_coo_gitconfig"
 write_coo_gitconfig
 
+# Validate BEFORE merging into ~/.claude/settings.json (#66): a
+# wrong-identity PAT must never land in the harness's persistent env
+# block. fetch_coo_secrets stages secrets in ~/.vade/coo-env and exports
+# them to this shell so validate_coo_identity can hit api.github.com;
+# only after that succeeds does merge_coo_settings_env write the PAT
+# into settings.json. set -e ensures we exit before merge on validate
+# failure.
 COO_BOOTSTRAP_STEP="validate_coo_identity"
 validate_coo_identity
+
+COO_BOOTSTRAP_STEP="merge_coo_settings_env"
+merge_coo_settings_env
 
 COO_BOOTSTRAP_STEP="summarize_coo_identity"
 summarize_coo_identity
--- a/scripts/coo-identity-digest.sh
+++ b/scripts/coo-identity-digest.sh
@@ -185,7 +185,7 @@ else
   echo ""
   echo "  WARN: identity is degraded — required env vars are missing."
   echo "  Inspect $BOOTSTRAP_LOG for the failing step, then re-run:"
-  echo "    VADE_FORCE_COO_BOOTSTRAP=1 bash \"${CLAUDE_PROJECT_DIR:-/home/user/vade-runtime}/scripts/coo-bootstrap.sh\""
+  echo "    VADE_FORCE_COO_BOOTSTRAP=1 bash \"${SCRIPT_DIR}/coo-bootstrap.sh\""
 fi
 
 # Classify OP_SERVICE_ACCOUNT_TOKEN visibility across build/session. The
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -596,6 +596,18 @@ GH_VERSION_DEFAULT="2.91.0"
 COO_AUTH_FP_EXPECTED="SHA256:9vxJc6c69L8eaR6CvwdZoYDco24W6yN6GkKwnsm8Uys"
 COO_SIGN_FP_EXPECTED="SHA256:pZeA8xycAtIsVGwhMzR3mg4KG05n9ksFuy4F1ZVXn3A"
 
+# Snapshot-persistent user bindir. Cloud harness runs as root and the
+# /home/user/ tree survives snapshot → resume; local Mac has no
+# /home/user/ and runs as the operator's user, so $HOME/.local/bin is
+# the right target. Both ensure_op_cli and ensure_gh_cli install here.
+_snapshot_user_bindir() {
+  if [ "$(id -u)" = "0" ] && [ -d /home/user ]; then
+    printf '/home/user/.local/bin'
+  else
+    printf '%s/.local/bin' "$HOME"
+  fi
+}
+
 ensure_op_cli() {
   # Install into a snapshot-persistent path so a build-time install is
   # still on disk at session-resume time. /root/ resets each resume in
@@ -605,7 +617,8 @@ ensure_op_cli() {
   # egress proxy is flaky (see run-2026-04-22T062313 and
   # run-2026-04-22T213126: "DNS cache overflow" 503 from the egress
   # gateway, both times).
-  local bindir="/home/user/.local/bin"
+  local bindir
+  bindir="$(_snapshot_user_bindir)"
   case ":$PATH:" in
     *":$bindir:"*) ;;
     *) export PATH="$bindir:$PATH" ;;
@@ -672,9 +685,8 @@ ensure_op_cli() {
 # Linux-only auto-install. On macOS (Darwin) the expectation is that
 # `brew install gh` has already satisfied check_cmd; if it hasn't, this
 # function refuses rather than dropping a non-runnable Linux binary
-# into ${HOME}/.local/bin. ensure_op_cli is currently Linux-only too
-# but hardcodes its bindir; aligning that pattern is tracked separately
-# so this PR stays scoped.
+# into ${HOME}/.local/bin. Bindir resolution is shared with ensure_op_cli
+# via _snapshot_user_bindir.
 #
 # Rationale: vade-app/vade-runtime#36 documents the mcp__github-coo__*
 # streamable-HTTP transport failure ("DNS cache overflow") that forces
@@ -686,11 +698,7 @@ ensure_op_cli() {
 # invariant stays load-bearing even when the primary MCP is down.
 ensure_gh_cli() {
   local bindir
-  if [ "$(id -u)" = "0" ] && [ -d /home/user ]; then
-    bindir="/home/user/.local/bin"
-  else
-    bindir="${HOME}/.local/bin"
-  fi
+  bindir="$(_snapshot_user_bindir)"
   case ":$PATH:" in
     *":$bindir:"*) ;;
     *) export PATH="$bindir:$PATH" ;;
@@ -839,12 +847,27 @@ fetch_coo_secrets() {
   chmod 600 "$env_file"
   log "  wrote $env_file (0600)"
 
-  _write_claude_settings_env "$github_pat" "$agentmail_key" "$mem0_key"
-
+  # Export to current shell so validate_coo_identity (which reads
+  # $GITHUB_MCP_PAT) sees the freshly-fetched PAT. settings.json mutation
+  # happens later in merge_coo_settings_env, only after validation
+  # passes — see #66 (env-merge-before-validate).
   [ -n "$github_pat" ]    && export GITHUB_MCP_PAT="$github_pat" GITHUB_TOKEN="$github_pat"
   [ -n "$agentmail_key" ] && export AGENTMAIL_API_KEY="$agentmail_key"
   [ -n "$mem0_key" ]      && export MEM0_API_KEY="$mem0_key"
   return 0
+}
+
+# Merge the (now-validated) COO env into ~/.claude/settings.json. Called
+# from coo-bootstrap.sh AFTER validate_coo_identity, so a wrong-identity
+# PAT never lands in the harness's persistent state. Reads from the
+# already-exported environment populated by fetch_coo_secrets — keeps
+# the call symmetric with the shell-export step there. See #66 for
+# the env-merge-before-validate failure mode this ordering closes.
+merge_coo_settings_env() {
+  _write_claude_settings_env \
+    "${GITHUB_MCP_PAT:-}" \
+    "${AGENTMAIL_API_KEY:-}" \
+    "${MEM0_API_KEY:-}"
 }
 
 # Merge COO env vars into ~/.claude/settings.json "env" object. Claude
@@ -1090,6 +1113,23 @@ validate_coo_identity() {
     return 1
   fi
   log "GitHub PAT valid for: $login"
+}
+
+# Quiet variant for the marker-shortcut precondition (#72): probes the
+# already-cached $GITHUB_MCP_PAT and returns 0 only when GitHub
+# confirms login=vade-coo. No retries (the marker shortcut is a fast
+# path; if the network is down, fall through to the full bootstrap
+# which has its own retry budget). No log output on the happy path —
+# only on fail-and-fall-through, so the operator sees why the marker
+# was bypassed. Single curl, hard 5s timeout — keeps the SessionStart
+# happy path within the existing budget envelope.
+_cached_pat_still_valid() {
+  [ -n "${GITHUB_MCP_PAT:-}" ] || return 1
+  local body login
+  body="$(curl -sfH "Authorization: Bearer ${GITHUB_MCP_PAT}" --max-time 5 \
+                  https://api.github.com/user 2>/dev/null)" || return 1
+  login="$(printf '%s' "$body" | grep -oE '"login"[[:space:]]*:[[:space:]]*"[^"]*"' | head -1 | sed -E 's/.*"([^"]*)"$/\1/')"
+  [ "$login" = "vade-coo" ]
 }
 
 summarize_coo_identity() {


vade-coo:
## Handoff prompt for next instance

> Verify vade-runtime#78 (bootstrap identity hygiene bundle) — `coo-bootstrap.sh`, `lib/common.sh`, `coo-identity-digest.sh` were modified. Cloud verification for #66/#72 + Mac verification for #39/#73 are gating the merge.
>
> **On a fresh CLOUD container after #78 lands:**
>
> 1. Confirm scripts updated:
>    `grep -c 'merge_coo_settings_env\|_cached_pat_still_valid\|_snapshot_user_bindir' /home/user/vade-runtime/scripts/lib/common.sh`
>    Expect: ≥3 (one per new function definition).
>
> 2. **#66 happy path:** the boot you're in already exercised the new ordering. Check `cat ~/.vade/coo-bootstrap.log | tail -5` — last line should be `OK step=complete rc=0`. Inspect `cat ~/.claude/settings.json | jq '.env | keys'` — expect `GITHUB_MCP_PAT`, `GITHUB_TOKEN`, `AGENTMAIL_API_KEY`, `MEM0_API_KEY` all present. If degraded, fail this step; investigate per `~/.vade/coo-bootstrap.log`.
>
> 3. **#66 fail-closed (manual):** ask Ven to temporarily put a wrong-identity PAT in `op://COO/vade-coo-self-2026-04/token`. Then `rm -f ~/.vade/.coo-bootstrap-done && VADE_FORCE_COO_BOOTSTRAP=1 bash /home/user/vade-runtime/scripts/coo-bootstrap.sh` — it must FATAL at `validate_coo_identity` AND leave `~/.claude/settings.json` env without `GITHUB_MCP_PAT` (run `cat ~/.claude/settings.json | jq '.env.GITHUB_MCP_PAT // empty'` — expect empty output) AND not create `~/.vade/.coo-bootstrap-done`. Restore the correct PAT and re-run to recover.
>
> 4. **#72 stale-PAT shortcut bypass:** with marker present and a *previously-valid PAT now revoked*, run `bash /home/user/vade-runtime/scripts/coo-bootstrap.sh` and confirm log says `marker stale (cached PAT failed validation); forcing re-run`. (If you can't revoke a PAT for this test, simulate by hand-editing `~/.claude/settings.json` to put a bogus token in `env.GITHUB_MCP_PAT`, then re-run.)
>
> 5. **#73 path resolver:** \`CLAUDE_PROJECT_DIR=/home/user bash /home/user/vade-runtime/scripts/coo-identity-digest.sh 2>&1 | grep VADE_FORCE_COO_BOOTSTRAP\` — must print the literal \`bash "/home/user/vade-runtime/scripts/coo-bootstrap.sh"\` (correct path), not \`/home/user/scripts/coo-bootstrap.sh\`. Trigger the remediation block by unsetting GITHUB_MCP_PAT in the digest's process: \`env -u GITHUB_MCP_PAT bash /home/user/vade-runtime/scripts/coo-identity-digest.sh | tail -20\` and look for the \`WARN: identity is degraded\` block.
>
> **On Mac (BLOCKING for #39 and #73 — must run on a Mac before clicking merge):**
>
> 6. **#39 — without brew op:** \`brew uninstall 1password-cli 2>/dev/null; rm -f ~/.local/bin/op\` then \`bash ~/GitHub/vade-app/vade-runtime/scripts/coo-bootstrap.sh\` (with a valid OP_SERVICE_ACCOUNT_TOKEN in env). Must NOT die on \`mkdir: Permission denied\` for \`/home/user/\`. After completion, \`ls -la ~/.local/bin/op\` and \`~/.local/bin/op --version\` should both succeed. Then \`brew install 1password-cli\` and re-run \`bash scripts/coo-bootstrap.sh\` — \`ensure_op_cli\` should short-circuit at \`check_cmd op\`.
>
> 7. **#73 on Mac:** \`unset CLAUDE_PROJECT_DIR; env -u GITHUB_MCP_PAT bash ~/GitHub/vade-app/vade-runtime/scripts/coo-identity-digest.sh | grep VADE_FORCE_COO_BOOTSTRAP\` — must print \`bash "/Users/<you>/GitHub/vade-app/vade-runtime/scripts/coo-bootstrap.sh"\` (or wherever the user's clone is). Then \`CLAUDE_PROJECT_DIR=/some/wrong/path env -u GITHUB_MCP_PAT bash ~/GitHub/vade-app/vade-runtime/scripts/coo-identity-digest.sh | grep VADE_FORCE_COO_BOOTSTRAP\` — same correct absolute path, regardless of \`CLAUDE_PROJECT_DIR\`.
>
> **On any failure:** check \`~/.vade/coo-bootstrap.log\` and \`~/.vade-cloud-state/build.log\` (cloud) or equivalent under \`~/.vade/local-state/\` (Mac). The trap in \`coo-bootstrap.sh\` records exit step + rc on every run. If the regression is not obvious, revert this PR rather than patch-on-top.
>
> **Followup blocked by #78:** none — Phase 1 of the audit is independent across the three PRs (#78, #79, #80). After #78 merges, no other queued work depends on it.
>
> **Greeting:** verify first, then greet with the bottom-line state-of-the-bundle.

vade-coo:
## Mac verification — #39 and #73

Ran on local Mac (Darwin arm64) against `claude/cloud-env-hardening-JjA7r-b1` at HEAD (commit `ee4dffe`).

### #73 — `coo-identity-digest.sh` path resolver — **PASS** (all three cases)

To trigger the remediation block reliably, ran with `env -i` and a clean `HOME` so neither process env nor `~/.claude/settings.json` contained `GITHUB_MCP_PAT` / `AGENTMAIL_API_KEY` (the existing process env + persisted settings.json otherwise short-circuit at the "Full identity loaded" branch).

| `CLAUDE_PROJECT_DIR` | Rendered path |
|---|---|
| unset | `bash "/Users/venpopov/GitHub/vade-app/vade-runtime/scripts/coo-bootstrap.sh"` |
| `/Users/venpopov/GitHub/vade-app/vade-runtime` (correct) | `bash "/Users/venpopov/GitHub/vade-app/vade-runtime/scripts/coo-bootstrap.sh"` |
| `/some/wrong/path` | `bash "/Users/venpopov/GitHub/vade-app/vade-runtime/scripts/coo-bootstrap.sh"` |

In every case the rendered path resolves to the actual sibling `coo-bootstrap.sh` (`SCRIPT_DIR`-based, immune to `CLAUDE_PROJECT_DIR` drift). `test -x` on the rendered path returns 0.

### #39 — `ensure_op_cli` bindir — **PASS for the stated symptom**, with one follow-up

**Test A — brew op pre-staged (current state):** PASS.

```
bash -c 'source scripts/lib/common.sh; ensure_op_cli'
→ [vade-setup] op CLI present: 2.30.3
```

`check_cmd op` short-circuits immediately; install branch never runs.

**Test B — op missing (simulated via `env -i HOME=$tmp PATH=/usr/bin:/bin:...` so brew is masked):**

- `_snapshot_user_bindir` → `$HOME/.local/bin` on Mac (was `/home/user/.local/bin`). ✅
- `mkdir -p "$bindir"` succeeds — **the original `mkdir: Permission denied for /home/user/` failure mode is gone**. ✅ This is the symptom #39 names.
- Download proceeds, `op` lands at `$HOME/.local/bin/op` (~24 MB).

**Follow-up surfaced (not in #39 scope, but adjacent):** `ensure_op_cli` lacks the Darwin-refuse guard that `ensure_gh_cli` has at lines 716–719. The install URL is hardcoded to `op_linux_${arch}_v${version}.zip`; on Mac the post-install probe correctly detects the broken install and returns 1 with `op CLI install appears broken`, but the function still leaves a 24 MB Linux ELF binary at `$HOME/.local/bin/op`:

```
file ~/.local/bin/op
→ ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, ...
~/.local/bin/op --version
→ cannot execute binary file
```

The PR's framing ("align bindir selection with `ensure_gh_cli`") closes the bindir half of the asymmetry. The OS-guard half is still open: on Mac without brew op, `coo-bootstrap.sh:108 ensure_op_cli` succeeds at `mkdir`, then drops a non-runnable Linux ELF, then returns 1. `local-setup.sh:119` is already gated by `[ "$(uname -s)" = "Linux" ]`, but `coo-bootstrap.sh` is not.

Recommend: file a follow-up issue for "ensure_op_cli: add Darwin-refuse guard symmetric to ensure_gh_cli" (one-liner: `case "$(uname -s)" in Darwin) log "op CLI not present on macOS; install via: brew install 1password-cli"; return 1 ;; esac` after the `check_cmd op` short-circuit). Not blocking for #78 — this PR closes the `/home/user/` symptom #39 names.

### Side-observation (#72 confirmation)

While running these tests I hit the `_cached_pat_still_valid` failure mode #72 closes: this session's `GITHUB_MCP_PAT` (loaded from a pre-PR-78 marker shortcut) is HTTP 401 against `api.github.com/user`, while a freshly-fetched PAT from `op://COO/vade-coo-self-2026-04/token` is HTTP 200. After #78 merges and the next bootstrap re-runs, the cached-PAT precondition will catch this on its own. (This comment is posted using the fresh 1P-fetched PAT.)

### Verdict

- #73: ready to merge.
- #39: ready to merge for the stated symptom; suggest opening a follow-up for the `ensure_op_cli` Darwin-refuse guard.
- #66 / #72: not Mac-verifiable; cloud verification is in your handoff prompt.

vade-coo:
Follow-up filed: vade-app/vade-runtime#81 — `ensure_op_cli` Darwin-refuse guard. Not blocking #78.

vade-coo:
## Post-merge handoff addendum (2026-04-25)

Two refinements caught during pre-merge probing on the open PR; they don't change the substance of the handoff above but will save the next verifier a head-scratch:

**1. Cloud-sha precondition.** Before running step 2's "happy path" check, confirm `~/.vade-cloud-state/integrity-check.json` `git_sha` is at or after `c61d260` (this PR's squash-merge commit). The cloud OS image is built per-container; if the boot you're in pulled an older snapshot, the bundle isn't actually live and step 2 would falsely confirm the OLD ordering. If `git_sha < c61d260`, stop and ask Ven to trigger a rebuild before continuing.

**2. Step 5 (#73) trigger correction.** `env -u GITHUB_MCP_PAT bash .../coo-identity-digest.sh` alone does NOT print the literal `bash "${SCRIPT_DIR}/coo-bootstrap.sh"` remediation line on a freshly-bootstrapped container. The digest's L181/L186 branches split on process-env vs settings.json-env: with PAT only unset in the calling process but still present in `~/.claude/settings.json`, the digest hits L181 ("env populated in settings.json but not in this process") — which is a different (and correct) warning, but it skips the L188 force-bootstrap echo. To trigger the L188 print you need both surfaces empty.

The structural #73 fix is verifiable statically without any env mutation: `grep -n SCRIPT_DIR /home/user/vade-runtime/scripts/coo-identity-digest.sh | head -3` should show `SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` at L15 and `${SCRIPT_DIR}/coo-bootstrap.sh` at L188. SCRIPT_DIR derives from BASH_SOURCE, not CLAUDE_PROJECT_DIR — so the printed remediation path can never drift relative to the digest's own location, regardless of CLAUDE_PROJECT_DIR. That is the root #73 contract.

Issue #111, opened 2026-04-26
cloud-setup: op build-time install still flakes after #76/#77; snapshots ship degraded


vade-coo:
## Problem

After #75/#76/#77 shipped (2026-04-25), the build-time `op` install **still flakes intermittently** and the snapshot ships with `op_installed_at_build=false`. Visibility (#77) and retry-budget tuning (#76) both work as designed — they're just not enough on their own. The session-side fallback then re-fetches `op` from `cache.agilebits.com` mid-SessionStart and is exposed to the same egress class the build-time install was supposed to absorb.

## Today's recurrence (2026-04-26 cloud session)

`/home/user/.vade-cloud-state/setup-receipt.json`:
```json
"built_at": "2026-04-26T20:26:00Z",
"op_installed_at_build": false
```

`/root/.vade/coo-bootstrap.log` (first two lines = SessionStart attempts):
```
2026-04-26T20:25:42Z START VADE_FORCE_COO_BOOTSTRAP=0
2026-04-26T20:26:00Z FAIL step=op_whoami rc=1
2026-04-26T20:26:01Z START VADE_FORCE_COO_BOOTSTRAP=0
2026-04-26T20:26:17Z FAIL step=op_whoami rc=1
```

Both SessionStart fallback attempts hit the same flake and gave up. Bootstrap halted; settings.json env never merged → all four PATs/keys (`GITHUB_MCP_PAT`, `GITHUB_TOKEN`, `AGENTMAIL_API_KEY`, `MEM0_API_KEY`) absent at MCP spawn → integrity-check Group D failed (D2–D6), and the Mem0 MCP spawned with empty `MEM0_API_KEY`, breaking `mcp__mem0__*` round-trips for the entire session until a manual `VADE_FORCE_COO_BOOTSTRAP=1 …/coo-bootstrap.sh` cleared everything (once `op` had stabilized, which it had by 20:28:21Z — same egress, ~2 min later).

Side-evidence the flake is generic egress, not 1Password-specific: in this same session, `mcp__github-coo__search_issues` returned `Streamable HTTP error: Error POSTing to endpoint: DNS cache overflow` — the very `undici` DNS-cache class that #36 was supposed to close for `github-coo` and that vade-runtime#110 just structurally retired for `mem0`.

## Why #77's surface alone wasn't enough

`#77` made the failure visible (receipt + SessionStart banner), so an operator/agent can *see* it. But:
[...]

Issue #112, opened 2026-04-26
zero-egress boot — close the cloud-boot flake chapter


vade-coo:
# Epic: zero-egress boot — close the cloud-boot flake chapter

## Why this epic exists

Over the past week, the cloud Claude Code boot pipeline has produced a rolling series of "boot is degraded" failures. Each individual fix has landed cleanly, yet the next session has surfaced a new variant. The whackamole pattern is empirical: predecessors include #36, #75, #76, #77, #109, #110, #111, plus MEMO-2026-04-23-03 / MEMO-2026-04-24-08 / MEMO-2026-04-26-16. Per BDFL: "this has literally been at the core of a lot of our problems in the past week … I want this chapter to end so we can start building great things."

The pattern resolved in this session: the symptom space "boot is degraded" hides **four distinct mechanisms** that are themselves symptoms of **one architectural pathology** — a boot path with too many serial network egress dependencies.

The four mechanisms:

1. **`undici` DNS-cache overflow** (Node-internal; hits any `type: "http"` MCP). Most likely root: `nodejs/undici#3901`. No env var disables it. Mitigation: don't use `type: "http"` for local MCPs.
2. **`cache.agilebits.com` external egress flake** (1Password binary CDN). Causes `ensure_op_cli` build-time install to fail; snapshots ship degraded; SessionStart fallback hits the same window.
3. **`my.1password.com` auth-endpoint transient** (`op whoami`). Different host from #2; ~30s windows.
4. **MCP env-at-spawn-time chicken-and-egg**. MCP spec injects env once at spawn; no hot-reload primitive. If `coo-bootstrap.sh` populates settings.json after MCP spawn, the MCP holds an empty key for the session.

Sharpest finding from the internal scan: **#36 closed without the proposed transport switch ever shipping.** `vade-runtime/.mcp.json:19-25` still declares `github-coo` as `type: "http"`. The closing comment substituted "use `gh` as fallback" but the `.mcp.json` was never updated. The wires were never pulled.

## Architectural principle

**Build-time absorption + zero-egress boot + stdio-only MCP.**

1. Anything binary or static is baked into the runtime image at build. No `cache.agilebits.com`, no `github.com/cli/cli` releases, no PyPI resolves on the boot path.
2. Boot-time egress is reduced to a single auth round-trip; results fan out from a cached bundle.
3. All MCP transports are stdio. No `type: "http"` entries in `.mcp.json`.
[...]

PR #115, opened 2026-04-27
integrity-check: deepen E5 with real get_memories round-trip (closes #114)

### LINKED ISSUE #114: E5 probe is shallow — passes when api.mem0.ai is 503-flapping
## Symptom

The E5 invariant in \`scripts/integrity-check.sh\` only runs the MCP \`initialize\` round-trip. When \`api.mem0.ai\` is returning 503 ("DNS cache overflow", the \`undici\` class named in MEMO-2026-04-26-16 §retirement), \`initialize\` still succeeds (it doesn't hit the API), so E5 reads green even though every tool call (\`get_memories\`, \`search_memories\`, etc.) errors with \`Expecting value: line 1 column 1 (char 0)\`.

Repro from this session (2026-04-27 13:11 UTC):

\`\`\`
$ curl -sS "https://api.mem0.ai/v1/ping/" -w "%{http_code}\n"
DNS cache overflow
503

$ /home/user/.local/bin/mem0-mcp-server (initialize)  → succeeds
$ get_memories tool call                              → "Expecting value: line 1 column 1 (char 0)"
$ bash mem0-rest.sh list-memo-pointers                → HTTP 503
$ jq '.groups.E.E5.ok' integrity-check.json           → true     ← misleading green
\`\`\`

## Why it matters

\`summary.ok\` is the load-bearing single-line answer to "did boot land". A green E5 here lets a session enter a Mem0 outage thinking the layer is healthy — \`/memo-sync\` will fail at session-end without warning, and the operator finds out only when they try to write.

Last night's nightly briefing flagged E5 as failing for a different reason (probe got no JSON-RPC in 6s). That's the same upstream class surfacing as a timeout instead of a 503 — both are upstream Mem0/Cloudflare symptoms, both get a misleading-green probe today.

## Proposed fix

Extend the E5 probe to do a real tool call (e.g. \`tools/call\` with \`get_memories\` filtered to a known-cheap user_id, page_size=1). Pass = JSON result with no \`isError: true\`. Keep the 6s timeout but raise it to ~10s for Mem0's cold-start variance.

Optional: split E5 into E5a (transport up — initialize) and E5b (API reachable — round-trip tool call). Reads degraded transport vs degraded API as different colors.

## Workaround for now

When E5 is green but a Mem0 write fails:

\`\`\`
curl -sS -o /dev/null -w "%{http_code}\n" https://api.mem0.ai/v1/ping/
\`\`\`

503 → outage; defer \`/memo-sync\` to next boot.

Found while addressing flags from MEMO-2026-04-27 nightly briefing.

https://claude.ai/code/session_014XjxbWtqvBmtu2SpNqbK7f

vade-coo:
## Summary

- Extend the E5 probe in `scripts/integrity-check.sh` to make a real `tools/call get_memories` round-trip after the existing `initialize` + `tools/list` handshake. Verdict now requires both layers to pass.
- Raise the timeout 6 s → 10 s for Mem0 cold-start variance; bump `head` 3 → 4 to absorb the extra JSON-RPC response.
- Distinguish three failure modes in the detail string so the operator/agent gets a specific remediation pointer: no JSON-RPC at all (transport down), handshake failure, or `api.mem0.ai` unreachable (handshake ok but round-trip failed).
- CI fake-env path unchanged — E5 still skips when `VADE_CI_WORKSPACE_ROOT` or `VADE_BINDIR_OVERRIDE` is set.

Closes #114.

## Why now

`api.mem0.ai` has been 503-flapping (\"DNS cache overflow\", the same `undici` class MEMO-2026-04-26-16 named) intermittently for a couple of days. The old E5 only exercised `initialize`, which succeeded against the local stdio binary even when the upstream API was down. `summary.ok` read green while every real Mem0 tool call errored with `Expecting value: line 1 column 1 (char 0)`. The probe lied; the session burned its first turns on empty memory loads.

Verified live this session: `curl https://api.mem0.ai/v1/ping/` → 503 `DNS cache overflow`. Before this PR, integrity-check reported 21/21 OK with E5 green. After this PR, integrity-check reports 20/21 DEGRADED with E5 detail `\"api.mem0.ai unreachable: handshake ok but get_memories round-trip failed — isError: Error executing tool get_memories: Expecting value: line 1 column 1 (char 0). Probe: curl https://api.mem0.ai/v1/ping/ ; if 503/DNS-cache-overflow, defer Mem0 writes (memo-sync, identity loads) until upstream recovers.\"`.

That is exactly the misleading-green case the issue names.

## Cheap-filter rationale

The round-trip uses `{filters: {AND: [{user_id: \"coo\"}]}, page: 1, page_size: 1}`. The COO namespace has ~10 records (CB-* and OG-*), so this is a tiny query — one record fetched, one network round-trip. Same shape as the always-load identity fetch in CLAUDE.md §4, so we know it's a path Mem0 is exercising in normal session use.

## Out of scope

- Splitting E5 into E5a (transport) + E5b (round-trip) as separate invariants. Worth doing if BDFL wants the colors split, but adds 1 invariant and the verdict ordering (handshake > round-trip > ok) already keeps the detail string specific. Defer to a follow-up if asked.
- Stream 2 binary baking for #111 — sibling PR.
- Self-host Mem0 — analyzed and deferred to a sibling sub-epic; filed alongside this PR.

## Test plan

- [x] Run `bash scripts/integrity-check.sh` while `api.mem0.ai` is 503 → E5 reports false with round-trip-failure detail; `summary.ok=false`.
- [ ] Run after `api.mem0.ai` recovers → E5 reports true with detail `stdio Mem0 MCP healthy: ... (handshake ok, X tools, get_memories round-trip ok)`.
- [ ] Bootstrap-regression CI (Layer-1) passes — E5 still skips in fake-env mode.
- [ ] Fresh-container post-merge confirmation (handoff comment to follow).

https://claude.ai/code/session_01XC1AF7LecoDXt5gMQAHhCf

### DIFF
--- a/scripts/integrity-check.sh
+++ b/scripts/integrity-check.sh
@@ -273,17 +273,30 @@ _add E2 skip "github-coo MCP retired by Epic #112 Stream 1; vade-coo identity ch
 _add E3 skip "requires-agent: inspect mcp-needs-auth-cache.json"
 _add E4 skip "requires-agent: observe tool namespaces"
 
-# E5: stdio Mem0 MCP server is installed and answers initialize.
-# The hosted https://mcp.mem0.ai endpoint hits a Node `undici` DNS-cache
-# overflow inside Claude Code's MCP HTTP transport on cloud-harness
-# boots (vade-runtime#36/#109), leaving the agent with no Mem0 surface
-# for the rest of the session. We bypass that by running the official
+# E5: stdio Mem0 MCP server is installed, answers initialize, AND
+# completes a real tool round-trip against api.mem0.ai. The hosted
+# https://mcp.mem0.ai endpoint hits a Node `undici` DNS-cache overflow
+# inside Claude Code's MCP HTTP transport on cloud-harness boots
+# (vade-runtime#36/#109), leaving the agent with no Mem0 surface for
+# the rest of the session. We bypass that by running the official
 # stdio MCP server (`mem0-mcp-server`, pinned via common.sh) as a
-# subprocess. This probe is the on-disk + handshake gate: if the binary
-# is missing OR the JSON-RPC initialize fails OR the tool list is
-# missing the read/write surface we depend on, mark E5 false. The
-# coo-identity-digest banner highlights any E* failure in the degraded
-# block, so a degraded Mem0 surface is loud at SessionStart.
+# subprocess.
+#
+# This probe layers two checks (per vade-runtime#114):
+#   1. Transport — the binary spawns and answers initialize +
+#      tools/list with the read/write tool surface we depend on.
+#      Fast failure mode: missing binary, no MEM0_API_KEY, JSON-RPC
+#      timeout, missing tool names.
+#   2. API reachability — a real tools/call get_memories round-trip
+#      that exercises api.mem0.ai. Without this, an
+#      api.mem0.ai 503 ("DNS cache overflow") leaves initialize
+#      passing while every actual tool call errors with
+#      "Expecting value: line 1 column 1 (char 0)" — a misleading
+#      green that defeats the point of `summary.ok`.
+#
+# Verdict ordering: handshake failure > round-trip failure > ok. The
+# coo-identity-digest banner highlights any E* failure in the
+# degraded block, so a degraded Mem0 surface is loud at SessionStart.
 #
 # Skip on CI (bootstrap-regression runs in fake-env mode against a
 # mock workspace; live MCP probes can't validate there) and when the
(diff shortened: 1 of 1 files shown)


vade-coo:

PR #115 deepens the E5 probe in `scripts/integrity-check.sh`. Boot-impacting because the integrity-check runs at SessionStart via `session-start-sync.sh`.

### Pre-merge gating

1. Check `gh pr checkout 115` and read the diff.
   - Expected: only `scripts/integrity-check.sh` touched, +65/-22 around the E5 block (lines ~276–390).
   - Failure response: if other files appear, abort and surface the divergence.

2. Run `bash scripts/integrity-check.sh` against current `api.mem0.ai` state.
   - If api.mem0.ai is healthy: expect `21/21 OK`, E5 detail `stdio Mem0 MCP healthy: /home/user/.local/bin/mem0-mcp-server (handshake ok, N tools, get_memories round-trip ok)`.
   - If api.mem0.ai is still 503-flapping: expect `20/21 DEGRADED — degraded (E5)`, E5 detail starting with `api.mem0.ai unreachable: handshake ok but get_memories round-trip failed`.
   - Either is acceptable for merge — the probe's job is to report ground truth.
   - Failure response: if E5 reports anything else (timeout, no JSON-RPC, handshake failure), the probe regressed; do not merge.

3. Probe `api.mem0.ai` directly to ground-truth the E5 verdict:
   ```
   curl -sS -o /dev/null -w "%{http_code}\n" https://api.mem0.ai/v1/ping/ --max-time 5
   ```
   - 200 → expect E5 green; 503 (or other 5xx) → expect E5 red. They must match.

### Post-merge confirmation

(fresh container from merged main)

1. Read `${VADE_CLOUD_STATE_DIR:-$CLAUDE_PROJECT_DIR/.vade-cloud-state}/integrity-check.json`.
   - Expected: `summary.ok` matches current api.mem0.ai health (curl-probe it).
   - If api.mem0.ai healthy: `summary.ok=true` and E5 detail mentions `round-trip ok`.
   - If api.mem0.ai degraded: `summary.ok=false` and E5 detail names the round-trip failure.
   - Failure response: if `summary.ok=true` but `mcp__mem0__get_memories` errors, the deepening regressed; reopen #114 with the verdict mismatch detail.

2. No further action required — the fix is deliberately read-only of behavior, not write.

### Followups not blocked

- Stream 2 binary baking (closes #111) lands on a sibling branch.
- Self-host Mem0 sub-epic filed under #112 alongside this PR.

Greet Ven before/after: after — this fix is observability-only, so any user prompt resumes the same conversation cleanly.

https://claude.ai/code/session_01XC1AF7LecoDXt5gMQAHhCf

Issue #109, opened 2026-04-26
Mem0 REST hit "DNS cache overflow" 503 at session start (#36 class extends to non-MCP egress)


vade-coo:
## Symptom

This session (run-2026-04-26T162126, cse_01QpF1Lxn6wXtYWeDCjYFyix), at SessionStart on a fresh post-Quorum-#6 cloud container:

- `github-coo` MCP failed to load (deferred-tools-removed reminder at start of turn 2; recovered by turn 3 — same #36 pattern).
- Mem0 MCP also disconnected at start.
- **Mem0 REST also unreachable**: `curl -s GET https://api.mem0.ai/v1/memories/?user_id=coo` returned HTTP 503 with body literally `DNS cache overflow`. Three retries 2s apart all 503'd. Self-recovered ~6 minutes later (HTTP 200, full payload).

Reproducible reading-side test that exhibited the failure:

```
$ curl -s -X GET "https://api.mem0.ai/v1/memories/?user_id=coo&page_size=50" \
    -H "Authorization: Token \$MEM0_API_KEY" -w "\n--- HTTP %{http_code} ---\n"
DNS cache overflow
--- HTTP 503 ---
```

After ~6 minutes of session activity (with no curl retries against api.mem0.ai), the next attempt returned HTTP 200 with all 17 records.

## Why this matters for #36's diagnosis

#36's root-cause section notes:

> **`mem0`** — also `type: http`, different CDN/endpoint behaviour. No overflow observed.

That observation no longer holds. Today's incident shows the same `DNS cache overflow` symptom on `api.mem0.ai`, which is:

- A different endpoint than `api.githubcopilot.com`
- Hit via direct `curl` (REST fallback per MEMO-2026-04-24-07), not via undici from inside the MCP server

So the failure mode extends beyond the MCP-server-via-undici path described in #36. The proxy or DNS-cache layer in the cloud sandbox is intermittently dropping resolutions for `api.*` hostnames generally, not just `api.githubcopilot.com` from inside an undici-DNS-cache path.

## Impact

- **Boot-pipeline degraded but not blocked**: per CLAUDE.md §6 graceful-degradation clause, fall back to `jq` over `coo/memo_index.json` + `grep coo/memos.md`. Worked fine — file-based grounding produced the diagnostic.
[...]

Issue #76, opened 2026-04-25
coo-bootstrap: tune retry budgets for transient 1Password egress hiccups


vade-coo:
## Problem

The shared `retry()` helper in `scripts/lib/common.sh:187-224` defaults to `tries=3` with exponential backoff `1s+2s = ~3s` of network tolerance. Two of the three SessionStart-critical 1Password egress paths inherit this default and have demonstrably been killed by transient hiccups that exceed the budget:

| Caller | Endpoint | Budget | Witnessed failure |
|---|---|---|---|
| `ensure_op_cli` (line 565) — `curl https://cache.agilebits.com/.../op_linux_*.zip` | Cloudflare-fronted CDN | `retry 3` (~3s) | run-2026-04-25T182206 — bootstrap died at `step=ensure_op_cli rc=1` after ~3s; egress recovered immediately after |
| `op_whoami` (`coo-bootstrap.sh:101`) — `op whoami` against `my.1password.com` / `api.1password.com` | 1P API | `retry 3` (~3s) | same session — `step=op_whoami rc=1` after ~6s of accumulated calls; one re-run succeeded |
| `_op_to_file` (line 942) — `op read ssh-key/private` | 1P API | `retry 5` (~15s) | run-2026-04-22T091701 needed 2-of-3 retries to clear cold-start; comment notes 5 was chosen specifically to clear that window |

`_op_to_file` is the one stage that's already been tuned up (5 attempts, 15s); the comment at lines 935–941 captures the lesson but it wasn't propagated.

## Why this matters

The egress flake is a recurring failure class — see prior in-source comments at `lib/common.sh:534-535` (run-2026-04-22T062313, run-2026-04-22T213126: \"DNS cache overflow\" 503 from the egress gateway). Today's session repeated the pattern. The cost of a tighter budget is a few seconds of bounded delay on the unhappy path; the cost of the current budget is a fully degraded SessionStart that requires manual `VADE_FORCE_COO_BOOTSTRAP=1` recovery and cascades into D2-D6+F4 integrity-check failures.

## Proposed fix

1. **Lift the `ensure_op_cli` curl retry budget to match `_op_to_file`** (5 attempts, ~15s tolerance). Same egress origin class, same flake pattern.
2. **Lift the `op_whoami` retry to 5 attempts** in `coo-bootstrap.sh:101`. Same flake mode.
[...]

PR #116, opened 2026-04-27
Stream 2: bake op + gh + mem0-mcp-server into Dockerfile; fail build on op flake (closes #111)

### LINKED ISSUE #111: cloud-setup: op build-time install still flakes after #76/#77; snapshots ship degraded

vade-coo:
## Summary

Three coordinated changes that close vade-runtime#111 and start epic vade-runtime#112 Stream 2.

1. **Dockerfile** bakes `op v2.31.0`, `gh v2.91.0`, and `mem0-mcp-server v0.2.1` into `/usr/local/bin` at image-build time. `cloud-setup.sh` ensure_*_cli functions already short-circuit on presence, so a runtime image using this Dockerfile pays the install cost once at image build instead of every snapshot rebuild. python3 + python3-venv added to the system-package layer for uv-managed mem0-mcp-server install.
2. **cloud-setup.sh** flips from WARN-and-continue to FATAL when `ensure_op_cli` fails. Closes #111 acceptance (a): the receipt is never written with `op_installed_at_build=false` because we exit before reaching the receipt-write step. The build.log FAIL line + absent receipt is the diagnostic.
3. **New CI workflow** `.github/workflows/docker-bake-check.yml` builds the Dockerfile and asserts the three binaries are present at `/usr/local/bin` and that pinned versions match `versions.lock`. Path filter is narrow (Dockerfile + versions.lock + the workflow file) so the docker-build cost only fires on image-level changes.

`versions.lock` adds the `mem0-mcp-server 0.2.1` entry for symmetry.

Closes #111. Advances #112 Stream 2.

## Why fail-build for op specifically (not gh / mem0-mcp-server)

`op` is a hard requirement: the COO identity load (`op read \"op://COO/vade-coo/credential\"`) cannot proceed without it, and Group D integrity-check invariants flip red without those credentials in `settings.json`. A snapshot without op is structurally guaranteed to fail the boot pipeline.

`gh` and `mem0-mcp-server` are degraded-not-fatal:
- Without `gh`, attribution falls through to venpopov (still works, just wrong identity on writes).
- Without `mem0-mcp-server`, the Mem0 surface goes dark (read-only fallback to `coo/memo_index.json` per CLAUDE.md §6).

Stream 2 can flip those to fail-build later if the BDFL wants symmetry, but the conservative interpretation of #111's acceptance is op-only.

## Trade-off (documented inline in cloud-setup.sh)

Cloud-build SLA goes down if cache.agilebits.com is flapping at snapshot-build time — today the snapshot ships degraded; with this PR the build fails. Two mitigations stack:

- The 5-attempt retry budget in `ensure_op_cli` (vade-runtime#76) absorbs short windows.
- The Dockerfile-baked `/usr/local/bin/op` layer in change (1) makes `ensure_op_cli` a no-op when the runtime image is in use — no fetch, no flake.

If neither is enough, the follow-up is #111 option (b) — alternate origin / local mirror.

## Honest scoping note

The cloud sandbox's base image is Anthropic-managed (the harness clones our repos in and runs `cloud-setup.sh` on top). Whether the cloud uses our `Dockerfile` directly is outside this PR's surface. Two interpretations:

- **If the cloud uses our Dockerfile**: changes (1) + (2) close #111 cleanly — the bake makes the fetch a no-op; fail-build is a safety net that should rarely fire.
- **If the cloud doesn't use our Dockerfile**: change (2) alone closes #111 by preventing degraded snapshots from shipping (at the cost of build-fail rate matching the cache.agilebits.com flake rate). The Dockerfile bake is groundwork for a future ops-level move.

Either way, change (3)'s Docker-bake CI catches Dockerfile drift on PR open.

## Test plan

- [x] Local syntax check: `bash -n cloud-setup.sh` passes; each Dockerfile RUN block parses under `bash -n`.
- [ ] `docker-bake-check` CI on PR open: builds image, verifies `op`, `gh`, `mem0-mcp-server` present at `/usr/local/bin` with pinned versions.
- [ ] `bootstrap-regression` CI on PR open: continues to pass (mocks satisfy `ensure_op_cli` regardless of fail-build flip).
- [ ] Fresh-container post-merge confirmation (handoff comment to follow): integrity-check 21/21 OK, `op_installed_at_build=true` in receipt, `op --version` matches pin.

## Out of scope

- Streams 3 (single-call credential bootstrap), 4 (full observability invariants A4/D7/E6 — PR #115 ships the E5 deepening separately), and 5 (doctrine memo).
- Symmetric fail-build for `gh` and `mem0-mcp-server` — degraded-not-fatal; defer.
- Self-host Mem0 — researched and filed as a sibling sub-epic under #112.
- Anthropic harness using our Dockerfile directly — ops-level coordination.

https://claude.ai/code/session_01XC1AF7LecoDXt5gMQAHhCf

### DIFF
--- a/.github/workflows/docker-bake-check.yml
+++ b/.github/workflows/docker-bake-check.yml
@@ -1,0 +1,82 @@
+name: Docker bake check
+
+# Layer-1 image-level CI for the vendored-binaries gate added by epic
+# vade-runtime#112 Stream 2. Builds the runtime Docker image from the
+# PR checkout and asserts the binaries that close vade-runtime#111 (op)
+# and that surface Mem0 + GitHub write paths (mem0-mcp-server, gh) are
+# baked into the image at /usr/local/bin. Catches Dockerfile drift like
+# "the layer was deleted" or "the version pin disagrees with versions.lock"
+# before merge, instead of catching it via a fresh-container boot
+# failure post-merge.
+#
+# Cheap path filter — only triggers when the Dockerfile or its pin
+# layer is actually touched. The bootstrap-regression workflow handles
+# script-level changes; this one handles image-level changes.
+
+on:
+  pull_request:
+    paths:
+      - 'Dockerfile'
+      - 'versions.lock'
+      - '.github/workflows/docker-bake-check.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: docker-bake-check-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  bake:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Build runtime image
+        run: docker build -t vade-runtime-test:${{ github.sha }} .
+
+      - name: Verify vendored binaries baked into the image
+        env:
+          IMG: vade-runtime-test:${{ github.sha }}
+        run: |
+          set -euxo pipefail
+          # Each binary must be present at a PATH location and runnable
+          # in the image. Closes vade-runtime#111 acceptance (a) by
+          # making the Dockerfile bake the substitute origin for `op`
+          # (and applying the same pattern to `gh` + `mem0-mcp-server`).
+          missing=0
+          for binary in op gh mem0-mcp-server; do
+            if ! docker run --rm "$IMG" sh -c "command -v $binary >/dev/null"; then
+              echo "::error::$binary missing from runtime image" >&2
+              missing=$((missing + 1))
+            fi
+          done
+          if [ "$missing" -gt 0 ]; then
+            exit 1
+          fi
+          # Version round-trip — `--version` for op and gh; mem0-mcp-server
+          # has no --version flag (we trust the install path test instead).
+          docker run --rm "$IMG" op --version
+          docker run --rm "$IMG" gh --version | head -1
+          docker run --rm "$IMG" sh -c "test -x /usr/local/bin/mem0-mcp-server"
+
+      - name: Verify pinned versions match versions.lock
+        env:
+          IMG: vade-runtime-test:${{ github.sha }}
+        run: |
+          set -euxo pipefail
+          # Read the pins from versions.lock (cheap inline parse — the
+          # file is short and the format is fixed). Compare against
+          # what the image actually ships.
+          op_pin="$(awk '/^op[[:space:]]+/ {print $2; exit}' versions.lock)"
+          gh_pin="$(awk '/^gh[[:space:]]+/ {print $2; exit}' versions.lock)"
+          op_actual="$(docker run --rm "$IMG" op --version 2>&1 | awk '{print $1}')"
+          gh_actual="$(docker run --rm "$IMG" gh --version 2>&1 | head -1 | awk '{print $3}')"
+          [ "$op_pin" = "$op_actual" ] || { echo "::error::op version drift: pin=$op_pin actual=$op_actual" >&2; exit 1; }
+          [ "$gh_pin" = "$gh_actual" ] || { echo "::error::gh version drift: pin=$gh_pin actual=$gh_actual" >&2; exit 1; }
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,8 @@ ARG USERNAME=node
 # build tools for any native npm deps, curl for debugging, procps for
 # tools like `ps` used by dev servers, openssh-client for git-over-ssh
 # and ssh-keygen fingerprint validation, unzip for the 1Password CLI
-# install path used by the COO identity bootstrap.
+# install path used by the COO identity bootstrap, python3 / python3-venv
+# for the uv-managed mem0-mcp-server install below.
 RUN apt-get update && apt-get install -y --no-install-recommends \
       git \
       ca-certificates \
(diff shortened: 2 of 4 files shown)


vade-coo:

PR #116 modifies the Dockerfile + cloud-setup.sh + adds a docker-bake-check workflow. Boot-impacting because cloud-setup.sh runs at snapshot-build time and a failure path now exits non-zero.

### Pre-merge gating

1. `gh pr checkout 116` and read the diff.
   - Expected: `Dockerfile` (+54), `versions.lock` (+9), `scripts/cloud-setup.sh` (+30/-5), `.github/workflows/docker-bake-check.yml` (new).
   - Failure response: if other files appear, abort and surface the divergence.

2. Watch the new `docker-bake-check` CI run on PR open.
   - Expected: green. Builds the Dockerfile, asserts `which op`, `which gh`, `which mem0-mcp-server` succeed in the image, and verifies `op --version` / `gh --version` match the pins in `versions.lock`.
   - Failure response: if the Dockerfile bake fails (cache.agilebits.com flake at CI time, uv install failure for mem0-mcp-server, etc.), the CI failure detail names which step. Most likely follow-up is to retry; if persistent, file a sibling issue.

3. Confirm `bootstrap-regression` CI still passes.
   - The fail-build branch in cloud-setup.sh fires only when `ensure_op_cli` returns non-zero. CI uses a PATH-shadowed `op` mock (`scripts/ci/mocks/op`) so `ensure_op_cli` succeeds and the failure path is not exercised.
   - Failure response: if bootstrap-regression goes red, check the integrity-check.json artifact uploaded by the workflow.

### Post-merge confirmation

(fresh container from merged main)

1. Read `${VADE_CLOUD_STATE_DIR:-$CLAUDE_PROJECT_DIR/.vade-cloud-state}/setup-receipt.json`.
   - Expected: `\"op_installed_at_build\": true` (the Dockerfile bake or the existing fetch path landed it).
   - If the receipt does not exist: the snapshot build failed — read `/home/user/.vade-cloud-state/build.log` for the FAIL line. Most likely cause is cache.agilebits.com 5xx flake during the build window (the new fail-build behavior). Remediation: trigger a snapshot rebuild; if the flake is sustained, file a follow-up.

2. Run `which op gh mem0-mcp-server` in the cloud session.
   - Expected: each binary resolves to a path. If `/usr/local/bin/...`, the runtime image baked it. If `/home/user/.local/bin/...`, the cloud harness's base image didn't include the Dockerfile and the cloud-setup.sh fetch path landed it (legacy mode; still works).
   - Failure response: if any binary is missing, integrity-check Group A/D will be red — investigate via the build.log.

3. Run `bash scripts/integrity-check.sh`.
   - Expected: 21/21 OK (or 20/21 if api.mem0.ai is 503-flapping per #114 — that's the PR #115 surface, not this PR's).
   - Failure response: if A1 or D-group invariants are red, the bootstrap chain didn't land cleanly.

### Followups not blocked

- Stream 3 (single-call credential bootstrap) and Stream 4 (full observability invariants).
- Symmetric fail-build for gh + mem0-mcp-server (currently WARN-and-continue).
- Anthropic harness using our Dockerfile directly — ops coordination, not code.

Greet Ven before/after: after — verification is mostly receipt + script reads; user prompt resumes cleanly.

https://claude.ai/code/session_01XC1AF7LecoDXt5gMQAHhCf

vade-coo (commented):
## Overview

Three coordinated changes:
1. **Dockerfile:** vendor `op` 2.31.0, `gh` 2.91.0, `mem0-mcp-server` 0.2.1 into `/usr/local/bin` at image-build time. Adds `python3` + `python3-venv` for the `uv tool install` of mem0-mcp-server.
2. **cloud-setup.sh:** flip `ensure_op_cli` failure from WARN-and-continue to FATAL (`exit 1`, no receipt). `gh` and `mem0-mcp-server` remain degraded-not-fatal.
3. **CI:** new path-filtered workflow asserts the three binaries exist at `/usr/local/bin`, and that op + gh versions match `versions.lock` (mem0-mcp-server has no `--version` flag, so presence-only).

The asymmetric fail-build (op only) is well-justified in both the PR body and the inline comment in `scripts/cloud-setup.sh`. Spot-checked: `MEM0_MCP_SERVER_VERSION_DEFAULT` is defined at `lib/common.sh:600` and `ensure_mem0_mcp_server` at `:832` — the version-pin comment in `versions.lock` is accurate.

## Strengths

- **Acceptance map is explicit.** The "deliberately not writing setup-receipt.json" comment ties the FATAL exit to MEMO-2026-04-26-17 / vade-runtime#111 acceptance (a) precisely. Downstream consumers can rely on absent-receipt-means-build-failed.
- **Path-filtered CI.** `paths:` correctly limits the docker-build cost to image-level changes; the existing bootstrap-regression covers script-level paths.
- **Self-disclosed trade-off.** The "Honest scoping note" in the PR body is exactly the kind of dual-interpretation flag needed to merge with eyes open.
- **Comments cite issues + memos.** Inline rationale references #76, #109, #111, MEMO-2026-04-24-08, MEMO-2026-04-26-16 — easy to bisect/audit later.

## Concerns

### Worth fixing before merge

1. **Unpinned `uv` install** (Dockerfile, mem0-mcp-server layer). `curl -fsSL https://astral.sh/uv/install.sh | sh` ships whatever `uv` is current at image-build time — the only unpinned binary in the entire layer. If a future uv breaks `uv tool install` semantics, image builds become non-reproducible and bisect-hostile, defeating the point of versioning the rest of the stack. Suggest `ARG UV_VERSION=...` and pull the matching release tarball, or pass `UV_VERSION=...` to the install script (it honors that env var).

2. **No retry budget on the uv install.** op and gh both use `curl -fsSL --retry 5 --retry-delay 2`. The uv installer has none — a single transient astral.sh blip fails the whole image. Same fix shape as (1).

3. **Architecture guard missing on `gh`.** The op layer fails fast on non-amd64/arm64 (`case ... esac`). The gh layer has no equivalent — non-supported arches fail with a silent 404 from URL composition. Cheap to harmonize.

### Quality-of-life (non-blocking)

4. **No buildx cache in docker-bake-check.** Every triggering PR pays the full apt + 3 binary fetches + npm globals. Adding `docker/setup-buildx-action` + GHA cache would shave subsequent runs significantly. Acceptable as-is given the narrow path filter.

5. **mem0-mcp-server version drift undetectable in CI.** No `--version` flag means the workflow can only check presence. Drift would have to come from a botched Dockerfile/versions.lock edit, which the `mem0-mcp-server==${MEM0_MCP_VERSION}` pin already prevents at build time. If you want CI symmetry, `pip show -V` against the uv-tools venv would do it.

6. **Absent-receipt as a signal is novel.** Worth confirming `session-start-sync.sh` / `coo-identity-digest.sh` handle the absent-receipt path explicitly — today they likely treat absent vs. present-with-`op_installed_at_build=false` as distinct degradations. Not strictly a PR-116 issue but the contract is now load-bearing.

## Risk summary

- **Build SLA:** PR raises build-fail rate to ≈ cache.agilebits.com flake rate when the cloud is *not* using our Dockerfile. Mitigations stack (5-retry + bake) but only the bake is structurally protective. Documented honestly; BDFL signal needed on the "honest scoping" interpretation.
- **Reproducibility:** the unpinned `uv` is a small but real soft spot in an otherwise-pinned image. Worth closing in this PR rather than as a follow-up.
- **Security:** no new exposure. All fetches HTTPS to vendor-canonical origins; binaries land in /usr/local/bin with `install -m 0755`. The `curl | sh` of the uv installer is the only piped-shell step — same risk profile as Anthropic's own install path, but pinning it would shrink the trust surface.

## Recommendation

**Lean approve with one small follow-up:** pin `uv` (concern 1), or open an issue for it before merge. Concerns 2–3 are worth folding in if you're touching the file anyway. Everything else is incremental.

> Posted as a comment-only review because GitHub blocks self-approval — PR author and reviewer are the same `vade-coo` identity. BDFL approval bit is yours to flip.

https://claude.ai/code/session_01WoRB67PbAC3KJWYNVzt2Zj

PR #122, opened 2026-04-27
auto-subscribe to PR activity after gh pr create

vade-coo:
## Summary

- New `scripts/auto-subscribe-pr.sh` PostToolUse Bash hook: detects `gh pr create`, parses the PR URL from tool output, and emits a system reminder telling the next turn to call `mcp__github__subscribe_pr_activity`.
- Wired into `.claude/settings.json` via the existing `vade-hooks/dispatch.sh` shim.

## Why this design

PR subscriptions are session-scoped to the model — a shell-side call cannot register one. The hook therefore injects `hookSpecificOutput.additionalContext` so the assistant issues the MCP call itself on the next turn. Removes the manual ritual of clicking "Auto-fix CI & address comments" on every PR.

## Test plan

- [x] Pipe-test (positive): synthetic `gh pr create` payload → emits proper `additionalContext` JSON with parsed `owner`/`repo`/`pullNumber`.
- [x] Pipe-test (negative, unrelated bash): silent no-op, exit 0.
- [x] Pipe-test (negative, `gh pr create --web`, no URL on stdout): silent no-op, exit 0.
- [x] Pipe-test (env-var-prefixed `GH_TOKEN=… gh pr create`): still detected.
- [x] End-to-end through `vade-hooks/dispatch.sh auto-subscribe-pr`: routes via `self_path` rule, returns correct JSON.
- [x] `jq` schema validation on settings.json passes.
- [ ] Post-merge: on a fresh container, run any `gh pr create` and confirm the assistant calls `subscribe_pr_activity` immediately after.

## Notes

- The hook fires on every Bash call but self-filters in ~1ms via a `case` match on the command string before doing any URL parsing.
- Always exits 0; misclassification cannot block the underlying tool.

https://claude.ai/code/session_01LJwyf263wgi2KKigyZ2nRJ

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -23,6 +23,14 @@
           { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-lifecycle --end" }
         ]
       }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" auto-subscribe-pr" }
+        ]
+      }
     ]
   },
   "env": {
--- a/scripts/auto-subscribe-pr.sh
+++ b/scripts/auto-subscribe-pr.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+# PostToolUse Bash hook: auto-subscribe to PR activity after `gh pr create`.
+#
+# Reads hook input JSON on stdin, detects whether the bash command was a
+# `gh pr create`, parses the PR URL from the tool response, and emits an
+# additionalContext system reminder instructing the next turn to call
+# mcp__github__subscribe_pr_activity. PR subscriptions are session-scoped
+# to the model, so a shell-side call cannot register one — the model must
+# issue the MCP call itself, hence the context-injection design.
+#
+# Always exits 0; no-ops cleanly on non-matching commands or unparsable
+# output so a misclassification can never block tool execution.
+
+set -uo pipefail
+
+input="$(cat 2>/dev/null || true)"
+[ -z "$input" ] && exit 0
+
+cmd="$(printf '%s' "$input" | jq -r '.tool_input.command // ""' 2>/dev/null || true)"
+case "$cmd" in
+  *"gh pr create"*) ;;
+  *) exit 0 ;;
+esac
+
+blob="$(printf '%s' "$input" | jq -r '.tool_response | tostring' 2>/dev/null || true)"
+url="$(printf '%s' "$blob" | grep -oE 'https://github\.com/[^/[:space:]"\\]+/[^/[:space:]"\\]+/pull/[0-9]+' | head -1 || true)"
+[ -z "$url" ] && exit 0
+
+owner="$(printf '%s' "$url" | sed -E 's|.*github\.com/([^/]+)/.*|\1|')"
+repo="$(printf '%s' "$url" | sed -E 's|.*github\.com/[^/]+/([^/]+)/pull/[0-9]+.*|\1|')"
+pn="$(printf '%s' "$url" | sed -E 's|.*/pull/([0-9]+).*|\1|')"
+
+if [ -z "$owner" ] || [ -z "$repo" ] || [ -z "$pn" ]; then
+  exit 0
+fi
+case "$pn" in *[!0-9]*|'') exit 0 ;; esac
+
+jq -n --arg owner "$owner" --arg repo "$repo" --argjson pn "$pn" '{
+  hookSpecificOutput: {
+    hookEventName: "PostToolUse",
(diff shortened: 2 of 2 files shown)


vade-coo:
Handoff prompt — verification on next container boot
=====================================================

Context
-------
PR #122 added a PostToolUse Bash hook (`scripts/auto-subscribe-pr.sh`) wired into `.claude/settings.json` that auto-subscribes the session to PR activity whenever `gh pr create` runs. New script + new hook entry; no other surfaces touched.

## Pre-merge gating

1. `gh pr checkout 122 -R vade-app/vade-runtime` and review the diff. There are exactly two files: `scripts/auto-subscribe-pr.sh` and `.claude/settings.json`. Anything else means a bad merge.
2. Run the pipe-test bundle:
   ```sh
   echo '{"tool_name":"Bash","tool_input":{"command":"gh pr create --title test"},"tool_response":{"stdout":"https://github.com/vade-app/vade-runtime/pull/231\n"}}' | bash scripts/auto-subscribe-pr.sh
   ```
   Expected: JSON on stdout with `hookSpecificOutput.additionalContext` containing `vade-app/vade-runtime#231`. Failure: any non-zero exit, or no JSON output.
3. Validate settings.json:
   ```sh
   jq -e '.hooks.PostToolUse[] | select(.matcher == "Bash") | .hooks[] | .command' .claude/settings.json
   ```
   Expected: exit 0, prints the dispatch invocation. Failure: exit 4/5 means a malformed entry.

## Post-merge confirmation

Required on a **fresh container booted from merged `main`** — `/root/.claude/settings.json` is copied at SessionStart by `vade-setup`, so the new hook only activates on next boot.

1. Verify the hook is loaded:
   ```sh
   jq '.hooks.PostToolUse' /root/.claude/settings.json
   ```
   Expected: shows the auto-subscribe-pr entry. Failure: settings sync regressed.
2. Run any `gh pr create` (or use a throwaway branch + draft PR if you don't have a real change pending). Expected: immediately after the bash call returns, the assistant calls `mcp__github__subscribe_pr_activity` with the parsed owner/repo/PR# without being asked. Failure: hook fires but assistant ignores the additionalContext — file an issue, the prompt wording may need to be more imperative.
3. Confirm CI/comment events flow back as `<github-webhook-activity>` blocks on the new PR.

## Followups blocked

None. Greet Ven before starting verification — the verification itself uses Ven's normal PR workflow, not a separate ritual.

https://claude.ai/code/session_01LJwyf263wgi2KKigyZ2nRJ

vade-coo:
## Review

Approve as-is; the fix here is the right shape (PostToolUse hook + `additionalContext` injection). vade-coo-memory#42 was the wrong surface — automated behaviors belong in hooks, not in CLAUDE.md rules the agent has to remember per turn. I'll close #42 separately and reference this PR.

### Pipe-test results

Ran the bundle from the PR description plus adversarials:

- Positive `gh pr create` with stdout URL → emits well-formed JSON ✓
- Unrelated bash → silent no-op ✓
- `gh pr create --web` (no URL) → silent no-op ✓
- `GH_TOKEN=… gh pr create` → still detected ✓
- `echo "gh pr create"` (substring false-positive) → silent no-op ✓
- Owner with shell-metachars in URL → silent no-op (regex rejects) ✓
- Empty input → silent no-op ✓
- `tool_response` as string instead of object → still works (good `tostring` choice) ✓

### Strengths

- `set -uo pipefail` (no `-e`, deliberately keeps going on jq parse hiccups)
- `jq -n --arg/--argjson` for safe JSON construction; no shell-injection surface
- Numeric validation on PR number (`case $pn in *[!0-9]*|'') exit 0`)
- Self-filters in <1ms via `case` match before any heavy parsing
- Always exits 0 — cannot block the underlying tool
- Routes through `hooks-dispatch.sh` self_path resolver, so it works on Mac (`CLAUDE_PROJECT_DIR`) and cloud (`mcp_symlink`) without code changes
- Stdin pipes through dispatch.sh cleanly (`bash "$RESOLVED_SCRIPT"` inherits stdin) — verified

### Two polishes I'm pushing as a follow-up commit

**(1) Tighten URL parsing to `tool_response.stdout` only.** Adversarial test:

\`\`\`json
{"command":"gh pr create --title x",
 "tool_response":{"stdout":"","stderr":"see https://github.com/foo/bar/pull/9 for example"}}
\`\`\`

Currently subscribes to `foo/bar#9` because `tool_response | tostring` slurps stderr too. Narrow real-world scenario, cheap to fix.

**(2) Drop "Do this silently in your next tool batch — do not narrate it."** Two reasons:

- Pushes the agent toward concealing a tool call from the user, which conflicts with the harness convention of stating tool intent before calling.
- A one-line "subscribed to vade-runtime#122 for CI/comments" is better UX than an invisible subscribe — that's how Ven knows the rule fired.

Replacing with a neutral instruction.

### "Always" question

The original CLAUDE.md PR (#42) raised whether subscribe-on-open should be unconditional. Hook-always is the right call here: subscriptions die with session, so a wasted subscribe on a session-exiting PR costs nearly nothing, and the hook architecture can't distinguish session intent anyway. Default-on is the only reachable design from this surface.

https://claude.ai/code/session_01MYvNUp5mAhVsnxZQkZQDVH

PR #128, opened 2026-04-28
binary-vendor: one-tarball release-asset path (briefing 004 / B1)

vade-coo:
## Summary

Replaces the four per-CDN fetches in `cloud-setup.sh` (op + gh + uv + mem0-mcp-server) with one curl from a private `vade-runtime` release asset using `\$GITHUB_MCP_PAT`. One github.com-class fetch, already-authenticated, snapshot-cache compatible.

Implements briefing [`vade-coo-memory/coo/briefings/004-cloud-binary-vendor.md`](https://github.com/vade-app/vade-coo-memory/blob/main/coo/briefings/004-cloud-binary-vendor.md) — but reframes the architecture from *private ghcr.io image + docker pull* to *private release asset + curl/tar*. The reframe is forced by a load-bearing finding from the read-first re-derivation: the cloud sandbox's docker daemon is non-functional at session time. `CAP_SYS_RESOURCE` is dropped from the kernel capability bounding set, blocking the privileged init path; rootless `dockerd` fails network-namespace + tap-device creation even with all userspace prereqs installed. Both blockers are sandbox-harness-level, not exposed in env config / settings.json / hooks. Detailed findings going to a follow-up issue this PR will reference.

`release-assets.githubusercontent.com` is in the cloud sandbox's Trusted-default allowlist alongside `ghcr.io`. Every structural benefit the briefing identified for ghcr.io (github.com class, already-authenticated, one artifact replaces four, snapshot-cache compatible) holds for the release-asset path with zero docker dependency.

## What's in the PR

- **`.github/workflows/publish-binary-vendor.yml`** — replicates the four Dockerfile install steps in an `ubuntu:24.04` container at the exact production layout (`/home/user/.local/{bin,share}/...`). Building at the production absolute path is necessary so the uv-managed mem0-mcp-server venv's shebangs resolve unchanged when the bundle is untarred on cloud. Tags `binary-vendor-<short-sha>` + `binary-vendor-latest`; publishes bundle + manifest + sha256 as release assets. Triggers on push to main when `Dockerfile` or `versions.lock` changes.

- **`scripts/lib/common.sh::ensure_binaries_from_vendor`** — cloud-only (root + `/home/user` gate). Idempotent short-circuit when all four binaries already present. Resolves asset_id via `api.github.com`, downloads via the octet-stream endpoint, sha256-verifies against `\$BINARY_VENDOR_BUNDLE_SHA256`, untars to `/home/user`. Best-effort: any failure returns 1 and the per-binary `ensure_*_cli` paths below run as fallback (briefing item 4 lean — keep direct-fetch fallback until release-asset reliability is measured in production).

- **`scripts/cloud-setup.sh`** — calls `ensure_binaries_from_vendor` before the existing `ensure_op_cli` / `ensure_gh_cli` / `ensure_mem0_mcp_server` block. Reads pin from `versions.lock`; treats `-` sentinel as unset.

- **`versions.lock`** — new `binary_vendor_bundle` line; `-` until the first `publish-binary-vendor` workflow run lands a SHA. Without a pin, the fetch path fails closed (release-lookup fails for non-existent tag) and falls through to per-binary fetch — the new path adds zero regression risk on the first merge.

- **`scripts/ci/run-bootstrap-regression.sh`** — sets `VADE_BINARY_VENDOR_DISABLE=1` so the regression suite's mock layer isn't extended for an endpoint that's never reached in CI.

## Posture decisions adopted (briefing items 2-4)

| Item | Briefing's lean | Adopted | Rationale |
|---|---|---|---|
| Visibility | Private | Private (vade-runtime release) | No EULA exposure for op; PAT auth already wired |
| Tag scheme | `:v<git-sha>` + `:latest` | `binary-vendor-<short-sha>` + `binary-vendor-latest` | Same shape, adapted for release tags |
| Fallback retention | Keep `ensure_*_cli` direct fetches | Kept | Defense-in-depth for first iteration; remove in a follow-up once measured |
| Bundle scope | 4 binaries | 4 (op, gh, uv, mem0-mcp-server) | Matches briefing scope; claude-code/tsx are npm-class with no observed flake |
| Host repo | (open) | vade-runtime | Same repo as `versions.lock` and the Dockerfile that drives the bundle; fewer moving parts than a dedicated repo |

## Bootstrap order (rollout)

This PR is the wire-up. The workflow auto-fires on the merge commit because the commit touches `versions.lock`. Sequence on merge:

1. PR merges → workflow fires → first bundle published as `binary-vendor-<sha>` + `binary-vendor-latest` with a SHA.
2. Follow-up PR pins `binary_vendor_bundle <sha>` in `versions.lock`. Next cache rebuild on cloud picks up the bundle path.
3. Until the pin lands, every cloud-setup.sh run logs a WARN ("binary vendor bundle unavailable") and falls through to per-binary fetch — same behavior as today, just with one more log line.

## Test plan

- [x] `bash -n` syntax check on all modified shell scripts
- [x] YAML lint on `publish-binary-vendor.yml`
- [x] Live smoke: `VADE_BINARY_VENDOR_DISABLE=1` early-returns
- [x] Live smoke: missing `\$GITHUB_MCP_PAT` returns 1 (no PAT branch)
- [x] Live smoke: idempotent short-circuit when all four binaries already present
- [ ] `bootstrap-regression.yml` passes on this PR (CI will report)
- [ ] Post-merge: publish-binary-vendor.yml fires and lands a release asset
- [ ] Post-merge: follow-up pins SHA in versions.lock; next cache rebuild fetches the bundle

## References

- Briefing: [`vade-coo-memory#coo/briefings/004-cloud-binary-vendor.md`](https://github.com/vade-app/vade-coo-memory/blob/main/coo/briefings/004-cloud-binary-vendor.md)
- Parent epic: vade-runtime#112 (mechanism-2: binary egress)
- Generalizes: vade-runtime#119 (op-mirror proposal — bundled into a 4-binary artifact rather than per-binary)
- Adjacent: vade-runtime#117 (correcting comment posted re: the docker-daemon premise)
- Followup: vade-runtime#129 (docker-on-cloud_default investigation; not blocking this PR)
- Memo: [MEMO-2026-04-28-01](https://github.com/vade-app/vade-coo-memory/pull/216) (adoption record + case-law)
- Case-law: MEMO-2026-04-23-02 (degraded-primary-surface pattern: same identity, different wire)

https://claude.ai/code/session_01YTWGG6nkTvXmTmqatM7Q4Z

### DIFF
--- a/.github/workflows/publish-binary-vendor.yml
+++ b/.github/workflows/publish-binary-vendor.yml
@@ -1,0 +1,276 @@
+name: Publish binary vendor bundle
+
+# Builds a single tarball containing the four flake-prone binaries that
+# scripts/cloud-setup.sh fetches from third-party CDNs at cloud-snapshot
+# build time (op, gh, uv, mem0-mcp-server), and publishes it as a
+# GitHub Release asset on this repo. cloud-setup.sh's
+# ensure_binaries_from_vendor then fetches the bundle in one curl with
+# $GITHUB_MCP_PAT, replacing four per-CDN fetches with one
+# github.com-class fetch from an already-authenticated origin.
+#
+# Why a release asset, not a ghcr.io image (briefing #004 directionally,
+# B1 reframe): the cloud sandbox's docker daemon is non-functional at
+# session time (CAP_SYS_RESOURCE dropped, network-namespace creation
+# denied — see follow-up issue), so docker pull from ghcr.io can't run
+# in cloud-setup.sh. release-assets.githubusercontent.com is in the
+# Trusted-default allowlist alongside ghcr.io, so the structural
+# benefits the briefing identified (one artifact, github.com class,
+# PAT-already-authenticated, snapshot-cache compatible) all hold for
+# release assets without the daemon dependency.
+#
+# The same Dockerfile install steps are replicated here (in an Ubuntu
+# container, no docker-in-docker required) with absolute target paths
+# matching production layout (/home/user/.local/...), so mem0-mcp-server's
+# uv-managed venv shebangs resolve unchanged when the bundle is untarred
+# at cloud-setup.sh time.
+#
+# Tag scheme: binary-vendor-<short-sha> per-build + binary-vendor-latest
+# moving tag for cloud-setup.sh to fetch. SHA256 of the bundle is
+# pinned in versions.lock; ensure_binaries_from_vendor verifies before
+# install.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'Dockerfile'
+      - 'versions.lock'
+      - '.github/workflows/publish-binary-vendor.yml'
+  workflow_dispatch:
+    inputs:
+      tag_suffix:
+        description: "Override tag suffix (default: short SHA)"
+        required: false
+        default: ""
+
+permissions:
+  contents: write
+
+concurrency:
+  group: publish-binary-vendor
+  cancel-in-progress: false
+
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:24.04
+    timeout-minutes: 20
+
+    steps:
+      - name: Install build prereqs
+        run: |
+          set -eux
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            curl ca-certificates git unzip tar jq python3 python3-venv \
+            xz-utils
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Read pinned versions from versions.lock
+        id: versions
+        run: |
+          set -eux
+          read_pin() {
+            local key="$1"
+            awk -v k="$key" '$1 == k { print $2; exit }' versions.lock
+          }
+          OP_VERSION="$(read_pin op)"
+          GH_VERSION="$(read_pin gh)"
+          UV_VERSION="$(read_pin uv)"
+          MEM0_MCP_VERSION="$(read_pin mem0-mcp-server)"
+          test -n "$OP_VERSION"
+          test -n "$GH_VERSION"
+          test -n "$UV_VERSION"
+          test -n "$MEM0_MCP_VERSION"
+          echo "op=$OP_VERSION"   >> "$GITHUB_OUTPUT"
+          echo "gh=$GH_VERSION"   >> "$GITHUB_OUTPUT"
+          echo "uv=$UV_VERSION"   >> "$GITHUB_OUTPUT"
+          echo "mem0=$MEM0_MCP_VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Stage production layout under /home/user/.local
+        run: |
+          set -eux
+          # Production cloud layout: /home/user/.local/{bin,share}/...
+          # mem0-mcp-server is uv-installed and its venv shebangs encode
+          # the absolute install path, so we must build at the exact
+          # path the bundle will untar to on cloud.
+          mkdir -p /home/user/.local/bin
+          mkdir -p /home/user/.local/share/uv-tools
+
+      - name: Install op CLI
+        env:
+          OP_VERSION: ${{ steps.versions.outputs.op }}
+        run: |
+          set -eux
+          arch="$(dpkg --print-architecture)"
+          tmp="$(mktemp -d)"
+          curl -fsSL --retry 5 --retry-delay 2 \
+            "https://cache.agilebits.com/dist/1P/op2/pkg/v${OP_VERSION}/op_linux_${arch}_v${OP_VERSION}.zip" \
+            -o "$tmp/op.zip"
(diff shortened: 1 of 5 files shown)


vade-coo:
## Handoff prompt for verifying PR #128 post-merge

The verification of this PR's wire-up depends on the **publish-binary-vendor.yml** workflow firing on the merge commit and producing the first release-asset bundle. That can only be observed after merge.

### Pre-merge gating

1. **`bootstrap-regression` PASS** — already reported in the sticky comment above (16/16 invariants).
2. **`bake` (docker-bake-check)** — verifies the Dockerfile bakes match `versions.lock` pins. Must be green before merge. (REQUIRED before merging.)
3. **Diff review** — `gh pr diff 128 --repo vade-app/vade-runtime`. Spot-check the four touch points: `.github/workflows/publish-binary-vendor.yml`, `scripts/lib/common.sh::ensure_binaries_from_vendor`, `scripts/cloud-setup.sh` call site, `versions.lock` `binary_vendor_bundle -` line.

### Post-merge confirmation

This is the actually-load-bearing verification. Run on a fresh container from merged main, in this exact order:

1. **Watch publish-binary-vendor.yml fire on the merge commit.**
   ```sh
   GH_TOKEN="$GITHUB_MCP_PAT" gh run list --repo vade-app/vade-runtime --workflow publish-binary-vendor.yml --limit 3
   ```
   Expected: a run started by the merge commit with status `completed`/`success`. Failure mode: workflow doesn't fire — check the path filter (`Dockerfile|versions.lock|.github/workflows/publish-binary-vendor.yml`); the merge commit DOES touch versions.lock so it should match.

2. **Verify the release was published.**
   ```sh
   GH_TOKEN="$GITHUB_MCP_PAT" gh release view binary-vendor-latest --repo vade-app/vade-runtime
   ```
   Expected: shows the moving tag with three assets — `binary-vendor.tar.gz`, `binary-vendor-manifest.json`, `binary-vendor.sha256`. Failure: workflow ran but release-create step failed; check the workflow logs.

3. **Read the published SHA from the manifest** (this is the value to pin in versions.lock for the follow-up PR):
   ```sh
   GH_TOKEN="$GITHUB_MCP_PAT" gh release download binary-vendor-latest --repo vade-app/vade-runtime --pattern binary-vendor-manifest.json --output - | python3 -c 'import json,sys; print(json.load(sys.stdin)["bundle_sha256"])'
   ```
   Expected: a 64-char hex string. Save it for step 5.

4. **Sanity-pull the bundle and inspect** (no install — just verify shape):
   ```sh
   tmp=$(mktemp -d) && cd "$tmp"
   GH_TOKEN="$GITHUB_MCP_PAT" gh release download binary-vendor-latest --repo vade-app/vade-runtime --pattern binary-vendor.tar.gz
   tar -tzf binary-vendor.tar.gz | head -20
   ```
   Expected: paths under `.local/bin/` (op, gh, uv, uvx, mem0-mcp-server) and `.local/share/uv-tools/mem0-mcp-server/...`. Failure: bundle is empty or missing a binary — workflow's install steps regressed; inspect logs.

5. **Open the follow-up PR pinning `versions.lock`.** The pin is what activates the fetch path; without it, `ensure_binaries_from_vendor`'s no-pin branch fires and proceeds without verification, but the asset isn't found because the lookup uses `binary-vendor-latest` which exists. Wait — re-reading: with no pin, the function downloads but doesn't verify. To gate adoption properly, pin the SHA from step 3 in versions.lock.
   ```sh
   # Replace '-' with the SHA from step 3
   sed -i.bak "s|^binary_vendor_bundle  -|binary_vendor_bundle  <SHA>|" versions.lock
   ```
   Open as a one-line PR titled `versions.lock: pin binary_vendor_bundle to <short-sha>`.

6. **Confirm cloud cache rebuild fetches the bundle.** Force a cache-rebuild (touch the env's setup script in the Anthropic cloud env UI, or wait for the ~7-day expiry). On the next session boot, expect `[vade-setup] binary vendor: sha256 verified` and `[vade-setup] binary vendor: installed op gh uv mem0-mcp-server to /home/user/.local/bin` in the digest. Failure: log says `binary vendor: sha256 mismatch` or `release lookup failed` — the publish step didn't update the moving tag, or the pin doesn't match the published SHA.

### Followups blocked by this PR

- vade-runtime#119 reframe (close once the binary_vendor_bundle pin lands and one cloud session has used the new path successfully).
- vade-runtime#117 docker-pull premise correction (a comment is being filed adjacent to this work).
- New issue scoping the docker-daemon-on-cloud_default investigation (filed as a sibling issue, not blocking this PR).

### Greet Ven

After step 6 confirms, greet Ven with: "binary-vendor B1 path adopted; first bundle published as `binary-vendor-<sha>`, versions.lock pinned in PR #<followup>, cloud cache rebuild verified the fetch+verify+untar path. Direct-fetch fallback retained per briefing item-4 lean. Next: measure release-asset reliability across N session rebuilds before removing the fallback."

https://claude.ai/code/session_01YTWGG6nkTvXmTmqatM7Q4Z

PR #157, opened 2026-04-28
sync: scope .claude config to workspace, stop touching $HOME/.claude on local

vade-coo:
## Summary

`local-setup.sh` and `session-start-sync.sh` were writing into the user's personal `$HOME/.claude/` on every run, overwriting their global Claude Code config (Warp plugin, autoMemoryEnabled, personal permissions). The root cause: `settings.json` hook commands hardcoded `$HOME/.claude/vade-hooks/dispatch.sh`, so the shim had to be installed at `$HOME` for the hook chain to resolve.

This PR switches the hook command path to `$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh`. Claude Code sets `CLAUDE_PROJECT_DIR` to the project root at hook-fire time — that's `$WORKSPACE_ROOT` on local and `$HOME` on cloud (where they're the same path). The shim already gets installed at `$WORKSPACE_ROOT/.claude/vade-hooks/` by `sync_claude_config`, so the workaround at `$HOME` is no longer needed.

## Changes

- `.claude/settings.json`: `$HOME` → `$CLAUDE_PROJECT_DIR` in all 10 hook commands plus the matching `Bash(...dispatch.sh:*)` permission entry.
- `scripts/local-setup.sh`: drop the redundant `ensure_hooks_dispatch_shim "$WORKSPACE_ROOT/vade-runtime/.claude" "$HOME/.claude"` call. The shim is already installed at `$WORKSPACE_ROOT/.claude/vade-hooks/dispatch.sh` by the prior `sync_claude_config` call.
- `scripts/session-start-sync.sh`: pass `$WORKSPACE_ROOT_DERIVED/.claude` as the explicit dst to both `sync_claude_config` and `aggregate_workspace_claude_config`. Previously these defaulted dst to `$HOME/.claude`, which silently overwrote global settings on every session start.

## Cloud impact

None. On cloud `$HOME == $WORKSPACE_ROOT == /home/user`, so all three target paths resolve to the same directory.

## Local impact

`$HOME/.claude/settings.json` (the user's personal Claude Code config) is no longer touched by the COO setup pipeline. The workspace `.claude/` at `$DEV_DIR/vade-app/.claude/` becomes the sole project-scope Claude Code config surface.

## Follow-up — other scripts still hardcode \$HOME/.claude

Out of scope for this PR but should be tackled next:
- `scripts/coo-bootstrap.sh` — writes API keys to `$HOME/.claude/settings.json` env block
- `scripts/persist-permissions.sh` — writes to `$HOME/.claude/settings.local.json`
- `scripts/integrity-check.sh` — checks `$HOME/.claude/settings.json` and `$HOME/.claude/vade-hooks/dispatch.sh`
- `scripts/session-idle-watchdog.sh` — reads `$HOME/.claude/projects/`
- `scripts/session-lifecycle.sh` — uses `$HOME/.claude/plans/`

## Test plan

- [ ] On local: launch Claude from `vade-app/`, confirm `$HOME/.claude/settings.json` is NOT modified by SessionStart.
- [ ] On local: confirm `vade-app/.claude/settings.json` reflects the latest `vade-runtime/.claude/settings.json` after SessionStart hook fires.
- [ ] On local: confirm SessionStart hooks (memo-index, coo-identity-digest, etc.) still fire correctly via the new \$CLAUDE_PROJECT_DIR path.
- [ ] On cloud: confirm no behavior change (paths are equivalent).

## Post-merge confirmation

After merging, the next fresh SessionStart on local should:
1. NOT touch `$HOME/.claude/settings.json` (verify via mtime).
2. Update `vade-app/.claude/settings.json` with the latest hook commands using \$CLAUDE_PROJECT_DIR.
3. Have working hook chain (verify integrity-check or look for boot.log entries).

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,26 +6,56 @@
       {
         "matcher": "startup",
         "hooks": [
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-start-sync" },
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" coo-bootstrap" },
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" memo-index" },
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" coo-identity-digest" },
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" discussions-digest" },
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-idle-watchdog --start" }
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" session-start-sync"
+          },
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" coo-bootstrap"
+          },
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" memo-index"
+          },
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" coo-identity-digest"
+          },
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" discussions-digest"
+          },
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" session-idle-watchdog --start"
+          }
         ]
       },
       {
         "hooks": [
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-lifecycle" }
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" session-lifecycle"
+          }
         ]
       }
     ],
     "Stop": [
       {
         "hooks": [
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-end-transcript-export" },
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" session-lifecycle --end" },
-          { "type": "command", "command": "bash \"$HOME/.claude/vade-hooks/dispatch.sh\" persist-permissions" }
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" session-end-transcript-export"
+          },
+          {
+            "type": "command",
+            "command": "bash \"$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh\" session-lifecycle --end"
+          },
+          {
+            "type": "command",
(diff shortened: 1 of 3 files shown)


vade-coo:

vade-runtime#157 modifies `.claude/settings.json` hook commands and the SessionStart sync targets — this is boot-impacting per CLAUDE.md and needs verification on a fresh container after merge.

### Pre-merge gating

\`\`\`bash
gh pr checkout 157 -R vade-app/vade-runtime
# Verify settings.json hook commands now reference \$CLAUDE_PROJECT_DIR (not \$HOME)
grep -c 'CLAUDE_PROJECT_DIR/.claude/vade-hooks' .claude/settings.json
# Expected: 11 (10 hook commands + 1 permission entry)

grep -c '"\$HOME/.claude/vade-hooks' .claude/settings.json
# Expected: 0
\`\`\`

If both probes match, the diff is correctly scoped and you can merge.

### Post-merge confirmation

On a fresh container from merged main:

\`\`\`bash
# 1. Confirm SessionStart hooks fired and resolved via the new path
tail -20 ~/.vade/boot.log
# Expected: session-start-sync entries with no errors
\`\`\`

\`\`\`bash
# 2. Confirm \$HOME/.claude was not modified by the sync (cloud only — on cloud HOME == WORKSPACE_ROOT, so this should still be touched, just via the workspace path resolution; this probe is mainly to confirm the local-side behavior holds elsewhere)
ls -la "\$CLAUDE_PROJECT_DIR/.claude/vade-hooks/dispatch.sh"
# Expected: symlink → vade-runtime/scripts/hooks-dispatch.sh
\`\`\`

\`\`\`bash
# 3. Sanity: integrity-check should not flag B4 (hooks drift)
jq '.summary, .groups.B.B4' "\${VADE_CLOUD_STATE_DIR:-\$CLAUDE_PROJECT_DIR/.vade-cloud-state}/integrity-check.json"
# Expected: summary.ok=true (or only known-OK degradations), B4.detail says "matches repo"
\`\`\`

### If the chain fails

If SessionStart hooks don't fire on local, the most likely cause is that \$CLAUDE_PROJECT_DIR isn't being set by Claude Code when launching from \$DEV_DIR/vade-app. Diagnose with:
\`\`\`bash
echo "CLAUDE_PROJECT_DIR=\${CLAUDE_PROJECT_DIR:-<unset>}"
\`\`\`
If unset, hook commands should fall back to a derived path. Open an issue on vade-runtime referencing #157 and revert.

Greet Ven before continuing follow-up work.

vade-coo:
Cloud regression flag, with a follow-up PR that fixes it: vade-runtime#158.

The "Cloud impact: None" claim assumes `\$HOME == \$WORKSPACE_ROOT == /home/user` on cloud, but integrity-check B5 in a live cloud session reports otherwise:

```
CLAUDE_PROJECT_DIR=/home/user cwd=/home/user HOME=/root
```

So at hook-fire time on cloud, `\$HOME` and `\$CLAUDE_PROJECT_DIR` resolve to **different** directories. Today, `cloud-setup.sh` only installs the dispatch shim under `\$HOME/.claude/vade-hooks/dispatch.sh` (= `/root/.claude/...`). After this PR, hook commands look for `/home/user/.claude/vade-hooks/dispatch.sh` — which doesn't exist on a fresh cloud snapshot — so every `SessionStart` hook (starting with `session-start-sync`, which would otherwise self-heal the layout) would fail file-not-found, leaving cloud sessions with no SessionStart hooks running.

The bootstrap-regression CI didn't catch it because in CI `CLAUDE_PROJECT_DIR=<unset>` (visible in the green check above) and the B4 invariant only validates settings.json shape, not whether the referenced shim path resolves at hook-fire time.

vade-runtime#158 adds a single extra `ensure_hooks_dispatch_shim` call in `cloud-setup.sh` to mirror the shim under `\$WORKSPACE_ROOT/.claude` alongside the existing `\$HOME/.claude` install. It's a no-op on local and pre-this-PR; it becomes load-bearing once this PR lands.

Recommended sequencing: merge #158 first (or together), then this PR. If this PR merges alone, snapshots built between this and #158 will boot with no SessionStart hooks on cloud.

https://claude.ai/code/session_01GtzT8KSUZyExkw51cMc2NW

PR #189, opened 2026-05-01
bootstrap: persist R2_TRANSCRIPTS_* + age identity in settings.json env

vade-coo:
## Summary

Extend `merge_coo_settings_env` to write the 3 R2/age secrets that `fetch_coo_secrets` already fetches from 1Password into `~/.claude/settings.json` env block:

- `R2_TRANSCRIPTS_ACCESS_KEY_ID`
- `R2_TRANSCRIPTS_SECRET_ACCESS_KEY`
- `TRANSCRIPTS_AGE_IDENTITY`

## Why

These three vars exist in `/root/.vade/coo-env` (written by `fetch_coo_secrets`) but the merge into `settings.json` only carried 4 of the 8 secrets. Tool subprocesses (Bash / Skill / sub-agents) only inherit `settings.json` env, so they had no R2 credentials.

Concrete impact: the Night's Watch §1 transcript-sidecar pass has been silently blocked for 3 consecutive nights. Throttle item `r2_credentials_missing_night_env` is live as of vade-coo-memory#359.

vade-coo-memory#243 (closed) was filed on the assumption these vars were present in the nightly run-env; this PR makes that assumption true.

## Changes

- `scripts/lib/common.sh`:
  - `_write_claude_settings_env`: 3 new positional args; node-script branches that write each to the merged env when non-empty.
  - `merge_coo_settings_env`: pass the 3 vars through from the bootstrap shell.
- `scripts/coo-bootstrap.sh`:
  - `_settings_env_complete` required-list adds the 3 keys, so a stale bootstrap marker forces a re-run when they're missing.

## Verification

Isolated test (preserves live `~/.claude/settings.json`):

```sh
TMPHOME=$(mktemp -d) && (
  source /root/.vade/coo-env
  source /home/user/vade-runtime/scripts/lib/common.sh
  HOME="$TMPHOME"
  merge_coo_settings_env
  jq '.env | keys' "$TMPHOME/.claude/settings.json"
)
```

Pre-patch settings.json env has 4 secrets; post-patch has 7. Confirmed.

## Out of scope (filing separately)

`boto3` install in the cloud container image. The Night's Watch needs both env credentials AND the boto3 module to enumerate R2. This PR fixes the credential-pass; boto3 install belongs in `Dockerfile`.

## Test plan

- [x] Isolated bootstrap test (above) shows 7 secrets post-patch.
- [ ] Live verification on next fresh-boot with `VADE_FORCE_COO_BOOTSTRAP=1` (handoff comment incoming).
- [ ] First post-merge nightly produces transcript sidecars instead of skipping the §1 pass.

## Authority

vade-runtime CLAUDE.md "may do autonomously" includes drafting script changes + opening PRs. Merging to main requires explicit BDFL approval.

https://claude.ai/code/session_01AKJyJFURFkjVei7Y9SdEko

### DIFF
--- a/scripts/coo-bootstrap.sh
+++ b/scripts/coo-bootstrap.sh
@@ -82,6 +82,10 @@ _settings_env_complete() {
     try { cfg = JSON.parse(fs.readFileSync(process.argv[1], "utf8")) || {}; }
     catch { process.exit(1); }
     const env = cfg.env || {};
+    // Bootstrap-blocking required-list: every key here, when missing, force-re-runs
+    // the full bootstrap. VADE_AUTH_TOKEN is intentionally NOT in this list — it
+    // is passed through merge_coo_settings_env (best-effort) but feature-gated and
+    // may be empty in some environments without warranting a full re-run.
     const required = ["GITHUB_MCP_PAT", "GITHUB_TOKEN", "AGENTMAIL_API_KEY", "MEM0_API_KEY",
                       "R2_TRANSCRIPTS_ACCESS_KEY_ID", "R2_TRANSCRIPTS_SECRET_ACCESS_KEY",
                       "TRANSCRIPTS_AGE_IDENTITY",
--- a/scripts/coo-bootstrap.sh
+++ b/scripts/coo-bootstrap.sh
@@ -83,6 +83,8 @@ _settings_env_complete() {
     catch { process.exit(1); }
     const env = cfg.env || {};
     const required = ["GITHUB_MCP_PAT", "GITHUB_TOKEN", "AGENTMAIL_API_KEY", "MEM0_API_KEY",
+                      "R2_TRANSCRIPTS_ACCESS_KEY_ID", "R2_TRANSCRIPTS_SECRET_ACCESS_KEY",
+                      "TRANSCRIPTS_AGE_IDENTITY",
                       "VADE_CLOUD_STATE_DIR", "PATH"];
     for (const k of required) { if (!env[k]) process.exit(1); }
     // PATH content sanity: Claude Code does not shell-expand env values,
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -1238,7 +1238,10 @@ merge_coo_settings_env() {
     "${GITHUB_MCP_PAT:-}" \
     "${AGENTMAIL_API_KEY:-}" \
     "${MEM0_API_KEY:-}" \
-    "${VADE_AUTH_TOKEN:-}"
+    "${VADE_AUTH_TOKEN:-}" \
+    "${R2_TRANSCRIPTS_ACCESS_KEY_ID:-}" \
+    "${R2_TRANSCRIPTS_SECRET_ACCESS_KEY:-}" \
+    "${TRANSCRIPTS_AGE_IDENTITY:-}"
 }
 
 # Persist non-secret bootstrap-derived path state into ~/.claude/settings.json
@@ -1268,6 +1271,7 @@ merge_coo_settings_paths() {
 # exists, so this is a no-op outside the Claude cloud image.
 _write_claude_settings_env() {
   local pat="$1" agentmail="$2" mem0="$3" vade_auth_token="${4:-}"
+  local r2_access_key_id="${5:-}" r2_secret_access_key="${6:-}" age_identity="${7:-}"
   if ! check_cmd node; then
     log "Warning: node missing; skipping ~/.claude/settings.json env merge"
     return 0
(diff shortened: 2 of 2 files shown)


vade-coo:
## Pre-merge gating

This PR is bootstrap-impacting: the `_settings_env_complete` required-list grows by 3 keys, so any container with a stale bootstrap marker that lacks the new keys will force-re-run the bootstrap on next boot. That's the desired behavior, but it's worth a fresh-boot verification before merge.

To verify in a fresh container before merge, in a new Claude Code session:

```sh
# Confirm the merge function with the patched code:
TMPHOME=$(mktemp -d)
( source /root/.vade/coo-env
  source /home/user/vade-runtime/scripts/lib/common.sh
  HOME="$TMPHOME"
  merge_coo_settings_env
  jq '.env | keys' "$TMPHOME/.claude/settings.json"
)
# Expect: R2_TRANSCRIPTS_ACCESS_KEY_ID, R2_TRANSCRIPTS_SECRET_ACCESS_KEY,
# TRANSCRIPTS_AGE_IDENTITY all present in the keys array.

# Confirm the marker re-run path on a stale bootstrap:
VADE_FORCE_COO_BOOTSTRAP=1 bash /home/user/vade-runtime/scripts/coo-bootstrap.sh
jq '.env | keys' /root/.claude/settings.json
# Expect: 3 new keys present.
```

## Post-merge confirmation

In a new container after merge:

1. Boot integrity-check `summary.ok` should be true (or no new failure
   class introduced by this change).
2. `jq '.env | keys' /root/.claude/settings.json` shows the 3 new keys.
3. Tool subprocesses can now access R2 — sanity-check via:
   ```sh
   echo $R2_TRANSCRIPTS_ACCESS_KEY_ID | wc -c   # 33 (32 hex + newline)
   ```

## Companion follow-up: boto3 install

The Night's Watch §1 transcript-sidecar pass needs **both** the env credentials (this PR) **and** `boto3` available in the nightly's Python. This PR ships only the env-pass; boto3 install belongs in the cloud container image (`Dockerfile`) and will land in a separate vade-runtime PR. Until both land, the §1 pass remains blocked even with this merged.

https://claude.ai/code/session_01AKJyJFURFkjVei7Y9SdEko

vade-coo:
## Review — peer-COO read

**Verdict: ship it.** Tight scope, correct fix, follows existing pattern, handoff comment already posted. Two minor observations below; neither blocks merge.

### What it does
Three secrets that `fetch_coo_secrets` already pulls from 1Password (`R2_TRANSCRIPTS_ACCESS_KEY_ID`, `R2_TRANSCRIPTS_SECRET_ACCESS_KEY`, `TRANSCRIPTS_AGE_IDENTITY`) now ride through `merge_coo_settings_env` into `~/.claude/settings.json` `env`, so tool subprocesses inherit them. `_settings_env_complete` adds the same three to its required-list, forcing a re-bootstrap on stale containers.

### Things that look right
- **Pattern parity** (`scripts/lib/common.sh:1271-1322`): same positional-arg + conditional-`if (process.env.X)` shape as the existing 4 secrets. No new abstraction, no signature shuffle.
- **Concrete-impact claim is grounded**: the 2026-05-01 nightly briefing flags "R2 credentials absent (3rd consecutive night; throttled for §5)" — vade-coo-memory#359 throttle item is real, so this PR clears a known blocker, not a hypothetical one.
- **Authorship**: opened as `vade-coo` ✓ (CB-005, MEMO-2026-04-22-04).
- **Handoff comment already posted** — pre-merge gating + post-merge confirmation per `coo/operations/handoff-prompts.md`. Good.
- **Out-of-scope `boto3` correctly carved out**: env-pass and runtime module are independent failures; bundling them into one PR would muddle the bisect.

### Observations (non-blocking)

1. **`VADE_AUTH_TOKEN` asymmetry.** `merge_coo_settings_env` passes `VADE_AUTH_TOKEN` through (`scripts/lib/common.sh:1241`), and `_write_claude_settings_env` writes it conditionally (`:1314-1316`), but it is **not** in the `_settings_env_complete` required-list — only the 3 R2/age vars are added. If `VADE_AUTH_TOKEN` is intentionally optional (e.g., feature-gated, not present in every environment), fine — but the asymmetry is undocumented in `coo-bootstrap.sh:75-90`. Worth a one-line comment at `coo-bootstrap.sh:85` clarifying which of the 8 secrets are bootstrap-blocking and which are best-effort, so the next instance touching this list doesn't silently break the symmetry the other way.

2. **Boot-loop floor on rotated secrets.** Adding all three to `required` means: if any one of them ever goes missing from 1Password (rotation gap, vault perms drift, COO vault re-key) `_settings_env_complete` returns false on every boot → `_cached_pat_still_valid` short-circuit fails → full bootstrap re-runs every container start. Acceptable for stable infra secrets, but if R2/age credentials get rotated we now have a hard dependency on all three being non-empty in the vault before next-boot succeeds quickly. Probably fine; just noting the new failure-mode floor.

### What I'd verify post-merge
The handoff comment already covers the right sequence — fresh-boot `jq '.env | keys' /root/.claude/settings.json` shows the 3 keys, then a single Night's Watch §1 cycle that produces sidecars (not skips). The latter is the actual outcome that retires #359's throttle item.

Approving in spirit — merge is BDFL's call.

https://claude.ai/code/session_01E9ohbp1BoE8rb3EfR487Z6

Issue #230, opened 2026-05-08
Investigate 5/8 nightly transcript export missing from R2 (autonomous-container Stop-hook gap)


vade-coo:
## What was observed

The 2026-05-08 Night's Watch session ran 00:35–01:15 UTC, completed cleanly, and filed PR vade-app/vade-agent-logs#261 with its session log and the BACKFILLED sidecars for the 2026-05-07 nightly's session (e60e6d1e). The 5/8 nightly's OWN session ciphertext, however, is missing from R2: no object exists under `transcripts/2026/05/08/` for any session_id.

A direct R2 enumeration via `scripts/lib/list-r2-transcripts.py` shows the most recent ciphertext is `transcripts/2026/05/07/e60e6d1e-…` (LastModified 2026-05-07T00:58:45), the 5/7 nightly's own export. The 5/8 nightly's expected `transcripts/2026/05/08/<session_id>.jsonl.gz.age` never landed.

## Distinct from vrt#229

vrt#229 is about E6's *detection logic* over-flagging quiet days due to the in-flight-jsonl false-positive. This issue is the *actual pipeline failure* E6 was designed to catch — and would catch correctly once vrt#229's fix is in place AND interactive activity resumes.

## Forensic context

- **No `claude/auto-meta-*` branches in vade-agent-logs since 2026-04-28.** The Stop-hook auto-commit path (vade-runtime#148 Part A) appears non-functional for autonomous nightly containers since at least 2026-04-29. The 5/7 nightly's meta.json was recovered via `transcript-meta-backfill.py` reading R2 object metadata (carried `_recovered_from_object_metadata: true`) — confirms 5/7 lost the auto-commit step but kept the R2 upload. **5/8 lost the R2 upload too.**
- The export wrapper `scripts/session-end-transcript-export.sh` forks via `setsid -f` and block-waits up to `VADE_TRANSCRIPT_EXPORT_BUDGET_SEC` (default 20s) for the upload child to finish. The wait-with-detach pattern from vrt#198 was added to handle hosted-container teardown timing.
- Local logs at `$HOME/.vade/transcript-export-logs/<TS>-<PID>.log` would show whether the upload child crashed, timed out, or completed — but autonomous containers are ephemeral and these logs are gone.

## Likely cause class

One of:
1. **Container teardown raced the upload.** The 20s budget didn't cover boto3 import + R2 PutObject when the harness PG-killed and then the container was destroyed quickly.
[...]

PR #231, opened 2026-05-08
wire VADE_RUNTIME_DIR + fix E6 in-flight false-positive (#228, #229)

### LINKED ISSUE #228: Wire VADE_RUNTIME_DIR into nightly container env
## Why

The Night's Watch standing order at `coo/nightly_review_task.md` §1.a calls scripts via `"$VADE_RUNTIME_DIR/scripts/lib/list-r2-transcripts.py"`. The env var resolves to empty in the nightly container — confirmed over two consecutive runs (2026-05-07 nightly vcm#503 §1, 2026-05-08 nightly vcm#504 §1).

Scripts succeed via explicit-path fallback (`/home/user/vade-runtime/scripts/lib/...`) so this hasn't broken anything operationally yet. But:

- The standing order's call form will silently fail if a future Watch instance follows the spec literally.
- An agent path that depends on `$VADE_RUNTIME_DIR` being set is one regression away from breaking.

## Scope

Wire `VADE_RUNTIME_DIR=/home/user/vade-runtime` into the nightly-container env. Likely surfaces:

- Container image build (whatever provisions `/home/user/vade-runtime`)
- Or the SessionStart hook layer if env injection happens there
- Or the project setup script if it already sets other VADE_* vars

## Verification

After fix lands, a fresh nightly boot should show `echo "$VADE_RUNTIME_DIR"` resolving to a non-empty path matching the actual checkout. No code changes to the standing order needed once the var is wired.

## Refs

- vade-coo-memory#503 — 2026-05-07 nightly briefing (first surface)
- vade-coo-memory#504 — 2026-05-08 nightly briefing (recurrence confirmed)
- `coo/nightly_review_task.md` §1.a — the call site that depends on the var

## Out of scope

Auditing every other VADE_* env var for the same wiring gap. If discovered during this fix, file as a follow-up.

https://claude.ai/code/session_01K24FefoFybhRU95Kqcu7W1

### LINKED ISSUE #229: integrity-check.json refresh regression: summary.timestamp null + stale E6 reading
## What was observed

Boot-time integrity check reports E6 degraded with detail "1 local jsonls in last 24h under /root/.claude/projects but R2 has 0 objects (prefixes: transcripts/2026/05/07/, transcripts/2026/05/08/) — transcript-export pipeline silent." This recurs on quiet days where the only "recent local activity" is the in-flight session itself.

## Real diagnosis (corrected from initial filing)

The original framing claimed `summary.timestamp = null` was the bug. **That was wrong** — the integrity-check schema does not carry a `summary.timestamp` field; the timestamp lives at top-level `checked_at` and IS fresh. My boot-time `jq '.summary.timestamp'` returned null because the field doesn't exist by schema, not because the check was stale.

The actual root cause is in `scripts/integrity-check-e6-r2-absence.py`'s `_count_local_jsonls()`: it walks `~/.claude/projects/**/*.jsonl` filtering by mtime within `VADE_E6_WINDOW_H` hours. **The current session's jsonl is being written continuously**, so its mtime is always within the window — inflating `local_count` even when the only live local activity is the in-flight session whose Stop-hook export hasn't fired yet.

Walking R2 LastModified for `transcripts/2026/05/`:
- `046dc00b-…` LastModified `2026-05-06T16:53:48` — outside 24h window
- `e60e6d1e-…` LastModified `2026-05-07T00:58:45` — barely outside (9h before window start)
- No 5/8 ciphertext

So E6's `r2_count=0` is technically correct for the 24h window. The false-positive comes from `local_count=1` being the in-flight session.

## Fix shape

`scripts/integrity-check-e6-r2-absence.py` now excludes the in-flight session via two opt-in paths:

1. **`/proc`-based in-flight detection** (Linux). Walks `/proc/*/fd` to find jsonl files held open by any process; the harness's active session log is reliably caught. Robust because it doesn't depend on the env-var format matching the jsonl filename.
2. **`--exclude-session` flag**, defaulting to `$CLAUDE_CODE_SESSION_ID`. Skips jsonls whose stem matches the given id. Works when the env-var format matches the jsonl stem (UUID form mid-session); the harness sometimes exposes a different identifier at hook time (e.g. cloud `cse_*` form), in which case the `/proc` path is the fallback.

## Verification

This session reproduces the false-positive: `uv run --script scripts/integrity-check-e6-r2-absence.py --verbose` reports `local_count=1, r2_count=0` before the fix, `local_count=0, ok=true` after.

Bootstrap-regression CI is unaffected: the suite stages a sandbox `$HOME` with no session jsonls.

## Side-finding (separate concern)

The 5/8 nightly's own ciphertext is genuinely missing from R2 — distinct issue from this E6 detection-logic fix. Filed separately.

## Refs

- vade-coo-memory#503 / vade-coo-memory#504 (nightly briefings that surfaced the boot-time false-positive over multiple days)
- vade-runtime#181 / vade-runtime#198 (transcript-export saga; E6 was designed to catch the *real* pipeline-silence pattern these resolved)
- vade-runtime#201 (E6 invariant origin)

https://claude.ai/code/session_01K24FefoFybhRU95Kqcu7W1

vade-coo:
## Summary

Two issue closes in one PR — both are scoped, both have local-test evidence, both passed adversarial auditors (safety + emancipatory) PASS-WITH-NOTES.

- **#228** — Wire `VADE_RUNTIME_DIR=/home/user/vade-runtime` through the canonical `merge_coo_settings_*` helper chain so hook subprocesses and Night's Watch invocations inherit it. Mechanical pattern application of the existing `merge_coo_settings_state_dir` shape.
- **#229** — E6 was reporting a false-positive on quiet days because the in-flight session's jsonl inflated `local_count`. Two opt-in exclusion paths: `/proc`-based primary (Linux) + `--exclude-session` flag (default `$CLAUDE_CODE_SESSION_ID`). Verified locally: this session's E6 flips from `degraded` to `ok`.

The original #229 issue body misdiagnosed the bug as `summary.timestamp = null`; that field doesn't exist by schema. The body has been corrected with the actual root cause.

Closes #228
Closes #229

## Audit fold

- Emancipatory auditor (PASS-WITH-NOTES, S=2/E=1) flagged the inline comment in `session-start-sync.sh` for citing a VADE-internal artifact as the primary rationale. **Folded** in commit a9e0af9 — generalized the comment, kept #228 as evidence pointer.
- Safety auditor (PASS-WITH-NOTES) flagged that `--verbose` prints `session=<CLAUDE_CODE_SESSION_ID>` to stderr; not a PAT but worth masking if log routing changes. **Deferred** as future hardening (issue not filed; flag for future ops review).
- Both auditors raised broader follow-ups (generic-helper refactor `merge_settings_env_var(key)`, pattern-registry for tooling patterns); **out of scope** for this PR — tracked here for future agency-elevation work.

## Test plan

- [x] Bash syntax check (`bash -n`) on `common.sh` + `session-start-sync.sh`
- [x] Python compile-check on `integrity-check-e6-r2-absence.py`
- [x] Live reproduction: `uv run --script scripts/integrity-check-e6-r2-absence.py --verbose` flips from `local_count=1, ok=false` to `local_count=0, ok=true` after the fix
- [x] Helpers callable: `type merge_coo_settings_runtime_dir` resolves; `_write_claude_settings_runtime_dir` resolves
- [ ] **Pre-merge gating** (boot-impacting; see standalone comment): bootstrap-regression CI green
- [ ] **Post-merge confirmation** (next fresh-container session): `echo "$VADE_RUNTIME_DIR"` resolves non-empty in hook subprocess env; integrity-check.json shows E6 ok with no degraded jsonl false-positive on a quiet day

## Side-finding (separate concern)

Investigation surfaced that the 5/8 nightly's own ciphertext is genuinely missing from R2 — distinct from the E6 detection-logic fix. Filed as #230.

https://claude.ai/code/session_01K24FefoFybhRU95Kqcu7W1

### DIFF
--- a/scripts/integrity-check-e6-r2-absence.py
+++ b/scripts/integrity-check-e6-r2-absence.py
@@ -112,15 +112,79 @@ def _date_prefixes(now: datetime) -> list[str]:
     return list(dict.fromkeys([yesterday, today]))
 
 
-def _count_local_jsonls(projects_dir: Path, window_start: datetime) -> int:
-    """Count *.jsonl under projects_dir with mtime >= window_start."""
+def _in_flight_jsonls() -> set[str]:
+    """Return realpaths of *.jsonl files currently held open by any
+    process — i.e. the harness's in-flight session log(s).
+
+    Linux /proc-based; returns empty set on systems without /proc
+    (macOS, BSD). The check is bounded by the number of running
+    processes; integrity-check is short-lived so this is cheap.
+    """
+    open_paths: set[str] = set()
+    proc = Path("/proc")
+    if not proc.is_dir():
+        return open_paths
+    for pid_dir in proc.iterdir():
+        if not pid_dir.name.isdigit():
+            continue
+        fd_dir = pid_dir / "fd"
+        if not fd_dir.is_dir():
+            continue
+        try:
+            for fd in fd_dir.iterdir():
+                try:
+                    target = os.readlink(fd)
+                except OSError:
+                    continue
+                if target.endswith(".jsonl"):
+                    try:
+                        open_paths.add(str(Path(target).resolve()))
+                    except OSError:
+                        open_paths.add(target)
+        except (OSError, PermissionError):
+            continue
+    return open_paths
+
+
+def _count_local_jsonls(
+    projects_dir: Path,
+    window_start: datetime,
+    exclude_session: str = "",
+    in_flight: set[str] | None = None,
+) -> int:
+    """Count *.jsonl under projects_dir with mtime >= window_start.
+
+    The current session's jsonl is being written continuously (mtime always
+    fresh) but its Stop-hook export hasn't fired yet, so counting it
+    inflates local_count and produces a false E6 alarm on quiet days where
+    the only "recent local activity" is the in-flight session itself
+    (vade-runtime#229). Two exclusion paths:
+
+    1. `in_flight` (set of realpaths): primary, /proc-based; catches
+       any jsonl currently held open by any process (the harness's
+       active session log). Robust because it doesn't assume the env
+       var format matches the jsonl filename.
+    2. `exclude_session`: secondary, env-var-driven; skips jsonls
+       whose stem matches the given id. Useful when the Claude Code
+       env carries a CLAUDE_CODE_SESSION_ID that matches the jsonl
+       stem (UUID format), but unreliable when the harness exposes
+       a different identifier (e.g. cloud `cse_*`) at hook time.
+    """
     count = 0
     if not projects_dir.is_dir():
         return 0
     cutoff_ts = window_start.timestamp()
+    in_flight = in_flight or set()
     for path in projects_dir.rglob("*.jsonl"):
         try:
             if path.stat().st_mtime >= cutoff_ts:
+                if exclude_session and path.stem == exclude_session:
+                    continue
+                try:
+                    if str(path.resolve()) in in_flight:
+                        continue
+                except OSError:
+                    pass
                 count += 1
         except OSError:
             continue
@@ -144,6 +208,14 @@ def main() -> int:
         help="Path to the Claude Code per-project jsonl tree. Default "
         "~/.claude/projects (also via VADE_E6_PROJECTS_DIR).",
     )
+    ap.add_argument(
+        "--exclude-session",
+        default=os.environ.get("CLAUDE_CODE_SESSION_ID", "").strip(),
+        help="Session id (jsonl stem) to exclude from local_count. The "
+        "in-flight session's jsonl is being written continuously and "
+        "would otherwise inflate the count on quiet days. Defaults to "
+        "$CLAUDE_CODE_SESSION_ID. vade-runtime#229.",
+    )
     ap.add_argument("--verbose", action="store_true", help="Per-row detail to stderr.")
     args = ap.parse_args()
 
@@ -164,10 +236,22 @@ def main() -> int:
     now = datetime.now(timezone.utc)
     window_start = now - timedelta(hours=args.window_h)
 
-    local_count = _count_local_jsonls(projects_dir, window_start)
+    in_flight = _in_flight_jsonls()
+    local_count = _count_local_jsonls(
+        projects_dir,
+        window_start,
+        exclude_session=args.exclude_session,
+        in_flight=in_flight,
+    )
     if args.verbose:
+        excl_bits = []
+        if args.exclude_session:
+            excl_bits.append(f"session={args.exclude_session}")
+        if in_flight:
+            excl_bits.append(f"in_flight={len(in_flight)}")
+        excl = f" (excl {', '.join(excl_bits)})" if excl_bits else ""
         _stderr(
-            f"local jsonls in last {args.window_h}h under {projects_dir}: "
+            f"local jsonls in last {args.window_h}h under {projects_dir}{excl}: "
             f"{local_count}"
         )
 
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -40,6 +40,15 @@ VADE_CLOUD_STATE_DIR="${VADE_CLOUD_STATE_DIR:-/home/user/.vade-cloud-state}"
 VADE_BUILD_LOG="${VADE_CLOUD_STATE_DIR}/build.log"
 VADE_SETUP_RECEIPT="${VADE_CLOUD_STATE_DIR}/setup-receipt.json"
 
+# vade-runtime working tree path. The Night's Watch standing order at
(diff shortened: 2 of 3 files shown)


vade-coo:
## Pre-merge gating

This PR touches the `~/.claude/settings.json` env surface (vrt#228 wires `VADE_RUNTIME_DIR`). Verification before merge:

- Bootstrap-regression CI must run green. The suite stages a sandbox `\$HOME` and exercises `cloud-setup.sh` → `session-start-sync.sh` end-to-end. The new `merge_coo_settings_runtime_dir` call should be exercised by it; expected result: `\$VADE_RUNTIME_DIR` lands in the staged settings.json env and integrity-check passes.

If CI flags a regression on the new helper, fix locally and force-push; do not bypass.

## Post-merge confirmation

Next fresh-container session (interactive or scheduled):

1. **In-session:** `echo "\$VADE_RUNTIME_DIR"` should resolve to `/home/user/vade-runtime` (or whatever the host-stable path is). Empty string = wiring didn't take.
2. **In hook subprocess:** confirm via integrity-check.json's groups.D (settings.json env surface) — it lists which env keys are present. `VADE_RUNTIME_DIR` should appear.
3. **Night's Watch quiet-day verification (vrt#229):** the next nightly that runs on a true-quiet day should NOT report E6 degraded with the "1 local jsonls but 0 R2 objects" false-positive. The detection should still fire correctly when local_count > 0 *excluding the in-flight session*, e.g. when an interactive session ran during the day but failed to export.

Open thread (out of scope for this PR): the 5/8 export gap surfaced during the investigation (#230) — separate from #229's detection-logic fix; needs its own pipeline-side investigation.

https://claude.ai/code/session_01K24FefoFybhRU95Kqcu7W1

vade-coo:
## Post-merge confirmation — partial pass

Run from this session boot (cloud harness, 2026-05-08 ~11:36 UTC).

### #228 — `VADE_RUNTIME_DIR` wiring: ✅ pass

- Shell env: `VADE_RUNTIME_DIR=/home/user/vade-runtime`
- `/root/.claude/settings.json` `.env.VADE_RUNTIME_DIR=/home/user/vade-runtime`
- Boot log shows `[vade-setup]   merged VADE_RUNTIME_DIR into /root/.claude/settings.json` on every dispatch run
- `dispatch.sh` Rule 1 (`env_VADE_RUNTIME_DIR`) is now the active resolution rule for hook-time runtime lookups (see boot.log)

### Confirmation step #2 — D4 lists `VADE_RUNTIME_DIR`: ⚠️ caveat

D4 reports `ok=true`, but `scripts/integrity-check.sh` L228 hardcodes the required-keys list as `[GITHUB_MCP_PAT, GITHUB_TOKEN, AGENTMAIL_API_KEY, MEM0_API_KEY, VADE_CLOUD_STATE_DIR]` — `VADE_RUNTIME_DIR` is not in the list, so the wiring passes-by-omission rather than by assertion. No regression-guard. Filed as part of vade-app/vade-runtime#232.

### #229 — E6 quiet-day false-positive: ❌ resurfaced at hook time

Boot-time integrity-check (11:36:48) still reported E6 degraded with the exact false-positive shape:
> `1 local jsonls in last 24h under /root/.claude/projects but R2 has 0 objects (prefixes: transcripts/2026/05/07/,transcripts/2026/05/08/)`

Root cause is **session-id form drift between hook subprocess and tool subprocess** on cloud:

| Context | `CLAUDE_CODE_SESSION_ID` | jsonl filename |
|---|---|---|
| SessionStart hook (boot.log 11:36:20) | `cse_01YSESw3PMWfBXKBEVsdPWCM` | `a5495c1d-...jsonl` |
| Bash tool subprocess (now) | `a5495c1d-80f7-4fa1-bed6-b78dc4f8f273` | (matches) |

At hook time, `cse_*` ≠ jsonl UUID stem → `--exclude-session` filter doesn't match. And on cloud, **no process holds the jsonl fd open continuously** (harness opens-write-closes per append, verified via `/proc/*/fd` walk → 0 jsonl holders) → `_in_flight_jsonls()` returns empty. Both exclusion paths fail by construction.

The probe docstring lines 162–172 explicitly anticipated this case — "unreliable when the harness exposes a different identifier (e.g. cloud `cse_*`) at hook time."

When I re-ran the chain manually from a Bash tool subprocess (`bash dispatch.sh coo-identity-digest` with CCSI=UUID), E6 flipped to `ok=true, local_count=0, "no recent jsonls"`. So the **probe code is correct**; the gap is the env-form assumption.

Filed as **vade-app/vade-runtime#232** with three sketched fix shapes (cse_*/UUID translation, recency-based third path, UUID-regex fallback).

### Side observation

F6 is also degraded (`external-touch above-floor: [lineage] the-eight/README.md: 7.8d > 7d`) — unrelated to this PR, just an external-touch refresh signal.

https://claude.ai/code/session_01YSESw3PMWfBXKBEVsdPWCM

PR #310, opened 2026-05-22
ops(scripts): git-push-coo wrapper via GIT_ASKPASS shim (closes #281)

### LINKED ISSUE #281: git-push-coo wrapper: keep PAT out of argv/trace via GIT_ASKPASS shim
## Problem

When the local git-proxy push path fails (see [#279](https://github.com/vade-app/vade-runtime/issues/279)) AND the documented fallback wrapper also fails (see [#280](https://github.com/vade-app/vade-runtime/issues/280)) — both observed once this session, 2026-05-20 — the only working push path is the credential-inline URL form:

```text
git push https://vade-coo:<COO_PAT>@github.com/vade-app/<repo>.git <branch>:<branch>
```

That works, but it has a real downside: bash expands the env-var into the command line, the expanded URL — including the literal PAT — appears in any shell-trace output the harness captures, and that trace lands in the encrypted session transcript on R2.

In this session that happened **twice** (push for vcm#792, push for vcm#793), and each occurrence required a follow-on PAT rotation recommendation. The current substrate has no structural defense; agents fall back to whatever path actually works under live failure, and the only working path leaks.

## Proposed structural fix

A small `bin/git-push-coo.sh` (or `scripts/git-push-coo.sh` under vade-runtime, sibling to the existing `git-push-with-fallback.sh`) that:

1. Writes a temp executable to `$(mktemp -d)/askpass.sh` containing:
   ```sh
   #!/bin/sh
   case "$1" in
     Username*) echo "vade-coo" ;;
     Password*) printenv COO_PUSH_TOKEN ;;
   esac
   ```
2. Invokes `GIT_ASKPASS=<that-script> COO_PUSH_TOKEN=$<COO PAT var> git push https://github.com/vade-app/<repo>.git <ref>:<ref>` — note the URL has **no credential component**, so the PAT never enters argv or the env in a printable form, and the askpass script only reads it from env (also not in argv).
3. Cleans up the temp dir on exit.

That keeps the token entirely out of:
- `argv` of any git subprocess
- the verbose `git push` URL banner that prints to stderr
- any shell-trace harness capturing command lines

The `COO_PUSH_TOKEN` env var is still set on the parent shell during the call, but env-var contents don't appear in `set -x` trace output the way expanded argv does. (Trace output shows `$VAR` references resolved-to-values when they're in command lines; not when they're just in env.)

## Alternative considered: persistent credential helper

`git config --global credential.helper '!sh -c "echo username=vade-coo; echo password=$<TOKEN_VAR>"'` configured once at bootstrap. Pros: no per-call shim. Cons:
- The credential.helper config string itself appears in `.git/config` / `~/.gitconfig`, including the env-var reference. Less leaky than inlining, but still a footprint.
- Bootstrap doesn't currently set this up (per [#279](https://github.com/vade-app/vade-runtime/issues/279) diagnosis: `git config --get-all credential.helper` returns empty post-boot).
- Adds a discoverable artifact (the gitconfig entry); the per-call shim leaves nothing behind.

I'd lean per-call shim, but either shape is structurally fine. The point is: agents shouldn't be in a position where the only working push path leaks credentials.

## Relationship to companion issues

- [#279](https://github.com/vade-app/vade-runtime/issues/279) is about the proxy returning 401 silently. Until fixed, the fallback path is load-bearing.
- [#280](https://github.com/vade-app/vade-runtime/issues/280) is about the documented `git-push-with-fallback.sh` exiting silently. Until fixed, the fallback to the *fallback* (manual inline) is what agents land on.
- This issue is about ensuring that **whatever path agents fall back to doesn't leak credentials.** It's somewhat orthogonal to fixing 279 / 280 — if both get fixed and the proxy works again, we never hit this code path. But while either is broken, this is the defense layer.

If the fallback wrapper ([#280](https://github.com/vade-app/vade-runtime/issues/280)) gets fixed, that wrapper should adopt the askpass shim internally so its own push doesn't leak either.

## Priority

P2. Not blocking (proxy push currently works for most operations; the leak only happens under failure-path use), but the cumulative-exposure shape (twice in one session) is the failure mode this prevents.

## What I'm not asking for in this issue

- A credential-leak audit of the existing inlined-URL push attempts in this session's transcript. The COO PAT rotation handles that out-of-band.
- A redesign of the proxy auth contract (that's [#279](https://github.com/vade-app/vade-runtime/issues/279)).
- Universal mandate to use the new wrapper. Agents can still use raw `git push` for the proxy path; this is the fallback-specific wrapper.

https://claude.ai/code/session_01DWi5vs4hDPSYkxxRdhT9U8

---

## Closure gate

Closes when: **Track 3** (rotation skill) of the [secrets-management implementation](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/implementation-kickoff.md) lands — the `GIT_ASKPASS` shim is the Class III paste-back surface for the `/rotate-credential` skill's manual-only rotation path. Class D1 defensive design per the audit's [failure axis](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/reports/failure.md). Member of the git-push family (vrt#279, #280, #281, vade-coo-memory#898) grouped under vade-coo-memory#871. Sub-issue of vade-coo-memory#871; Project Status=Blocked.

PR #310 (closed 2026-05-23) attempted an isolated patch outside Track 3's design; per the triage, the wrapper lands as part of the skill, not as a standalone script.

_Per [triage row](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/triage.md). Enacted via vade-coo-memory#932._

vade-coo:
## Summary

Adds `scripts/git-push-coo.sh`, a small wrapper that pushes via a
credential-clean URL while supplying the PAT through a temporary
`GIT_ASKPASS` helper. Closes vade-runtime#281.

## Why

When the cloud git-proxy fails (vade-runtime#279) and the documented
fallback wrapper also fails (vade-runtime#280), agents reach for the
credential-inline URL form (see #281 issue body for the literal
shape). bash expands the env-var into the command line, so the
literal PAT appears in (a) every git subprocess argv (visible via
`ps aux`), (b) any shell-trace capture (`set -x`, harness xtrace),
and (c) the verbose git URL banner. Two sessions in a row this
triggered out-of-band PAT rotations (2026-05-20).

## Threat model and mechanism

The wrapper uses git's built-in `GIT_ASKPASS` mechanism:

1. Write a tiny askpass shell script to a `mktemp -d` directory
   (mode 0700, parent dir mode 0700).
2. The askpass script reads `COO_PUSH_TOKEN` and `COO_PUSH_USER` from
   env and echoes them to stdout in response to git's "Username for
   ..." / "Password for ..." prompts.
3. `export GIT_ASKPASS=<path>`, then `git push` against a clean URL
   (no credential component).
4. Any `https://user:pass@host/...` URL in argv is rewritten to
   `https://host/...` before invoking git, so the caller can't
   accidentally leak even if they pass the inline form.
5. `trap cleanup EXIT INT TERM HUP` removes the helper dir on every
   exit path.

`set -x` shows resolved values when env vars land in command lines,
not when they sit in env. So the PAT enters memory but no printable
trace stream.

`GIT_TERMINAL_PROMPT=0` is also set so git fails loud rather than
hanging on an interactive tty if askpass misbehaves.

## Verification

Local smoke-test against a bare repo (sandbox):

- Missing PAT env: wrapper prints diagnostic and exits 2.
- With the PAT env set and a local bare-repo remote: push succeeds,
  no PAT in argv (verified via a PATH-shadowed fake git that echoes
  its argv to stderr).
- Credential-inline URL input rewriting: argv printed by the fake
  git shows the clean URL form, with the user:pass segment stripped.
- After exit: no `git-askpass-coo.*` directories remain in `/tmp`
  (trap cleanup works).

## Test plan

- [ ] CI: bootstrap-regression workflow runs and stays green
      (this PR touches `scripts/`, so the workflow fires).
- [ ] Manual: BDFL or follow-up COO session uses the wrapper against
      a real `vade-app/*` repo to confirm end-to-end push works
      against `api.github.com`.

## Decisions deferred

- **Migration of `git-push-with-fallback.sh` to use this helper.**
  That wrapper currently constructs a credential-inline URL and
  relies on a sed redactor to mask output. Migrating is non-trivial
  because it composes args dynamically (strips `-u`, restores
  upstream tracking post-push) and selects PATs by owner. Left as a
  `TODO` comment in the new file; tracking as a follow-up to keep
  this PR S-sized.
- **Universal mandate to use the wrapper.** Per #281's "What I'm not
  asking for", the wrapper is the safe path for the credential-inline
  fallback — raw `git push` against the proxy stays the default.

https://claude.ai/code/session_011dTuSjsfYFazcEecjpsHKR

### DIFF
--- a/scripts/git-push-coo.sh
+++ b/scripts/git-push-coo.sh
@@ -1,0 +1,153 @@
+#!/usr/bin/env bash
+# git-push-coo — credential-safe direct-URL git push wrapper.
+#
+# THREAT MODEL
+# ------------
+# The cloud git-proxy occasionally returns HTTP 403 / drops connections
+# on push (vade-runtime#67, #279, #280). When the proxy path and the
+# documented fallback wrapper (`git-push-with-fallback.sh`) both fail,
+# agents under live failure typically reach for the credential-inline
+# URL form:
+#
+#   git push "https://vade-coo:${GITHUB_MCP_PAT}@github.com/<owner>/<repo>.git" \
+#     <branch>:<branch>
+#
+# That works, but bash expands $GITHUB_MCP_PAT into the command line.
+# The expanded URL — including the literal PAT — appears in:
+#   · argv of every git subprocess (visible in `ps aux`)
+#   · shell-trace output (`set -x`, harness xtrace captures)
+#   · the verbose `git push` URL banner that prints to stderr
+#   · any session transcript that captures those streams
+#
+# Two PAT exposures in a single session (vcm#792, vcm#793 pushes,
+# 2026-05-20) triggered out-of-band rotation recommendations. This
+# wrapper closes the leak structurally.
+#
+# MECHANISM
+# ---------
+# Git supports `GIT_ASKPASS` — a path to an executable that git
+# invokes to prompt for credentials. The askpass binary receives a
+# single argv: the prompt text ("Username for '<url>': " or
+# "Password for '<url>': "). Whatever it prints to stdout is the
+# answer. Critically:
+#   · The PAT is read from an env var inside the askpass helper, so
+#     it never appears in the helper's argv.
+#   · The git push URL has no credential component, so the PAT never
+#     appears in git's argv either.
+#   · Env-var contents don't appear in `set -x` trace output the way
+#     expanded argv does. (`set -x` shows resolved values when they
+#     land in command lines; not when they sit in env.)
+#
+# Net: the PAT enters memory but not any printable command-line trace.
+#
+# USAGE
+# -----
+#   GITHUB_MCP_PAT=<pat> scripts/git-push-coo.sh <git-push-args>...
+#
+# Example:
+#   scripts/git-push-coo.sh origin HEAD:refs/heads/my-branch
+#   scripts/git-push-coo.sh https://github.com/vade-app/foo.git main
+#
+# The wrapper rewrites a credential-bearing remote URL (if present)
+# to a clean URL before invoking git push; the PAT moves into the
+# askpass helper instead. If the remote arg is a symbolic name like
+# `origin`, the configured URL is used as-is (assumed clean).
+#
+# RELATIONSHIP TO EXISTING WRAPPERS
+# ---------------------------------
+# `scripts/git-push-with-fallback.sh` currently constructs a
+# credential-inline URL for its fallback push and relies on a sed
+# redactor to mask output. That wrapper should migrate to use the
+# askpass mechanism in this file — see TODO below. Migration is
+# non-trivial because the fallback wrapper composes args dynamically
+# (strips `-u`, restores tracking post-push) and the PAT selection
+# is owner-dependent (GITHUB_MCP_PAT vs GITHUB_PUBLIC_PAT). Tracking
+# follow-up issue rather than refactoring inline here.
+#
+# TODO(vade-runtime#281 follow-up): migrate `git-push-with-fallback.sh`
+# to source the askpass helper from this script (or factor the helper
+# into `scripts/lib/git-askpass.sh`). Keep the per-call shim shape;
+# don't switch to a persistent credential.helper (the gitconfig
+# footprint is itself an artifact, see #281 "Alternative considered").
+
+set -uo pipefail
+
+readonly PAT_ENV_DEFAULT="GITHUB_MCP_PAT"
+readonly COO_USER_DEFAULT="vade-coo"
+
+# Pick PAT env var: explicit COO_PUSH_TOKEN_ENV beats default. The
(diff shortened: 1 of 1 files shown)


venpopov:
not good. This does the antipattern of stacking a yet another layer and script on an already overcomplicated stack, and it completely ignores the security audit plan that just landed and the commissioned implementation that establishes procedures for secret management, env variables and so on. It's unclear whether the existing secrets plan already covers this, but even if not, this shouldn't be an isolated one-off patch like this

Issue #279, opened 2026-05-20
git push to local proxy returns 401 silently (observed-once)


vade-coo:
## Symptom

Single-session observation 2026-05-20 (run-2026-05-20T181315, cloud Claude Code container): `git push origin <branch>` exits rc=1 with **no stderr output**. Tried `-v`, `GIT_TRACE=1`, `GIT_CURL_VERBOSE=1` — all silent. Fetch was fine in the same session.

## Diagnostic

Direct curl against the local proxy bypassing git:

```
$ curl -v http://127.0.0.1:35033/git/vade-app/vade-coo-memory/info/refs?service=git-upload-pack
< HTTP/1.1 401 Unauthorized
< Www-Authenticate: Basic realm="Git Proxy"
< Content-Length: 0
```

So the proxy is rejecting unauthenticated requests. The remote URL has the username `local_proxy` but **no password component**:

```
remote.origin.url=http://local_proxy@127.0.0.1:35033/git/vade-app/vade-coo-memory
```

And no credential helper is configured:

```
$ git config --get-all credential.helper
(empty)
```

So git has no source for the Basic-auth password the proxy now wants. This worked for fetches in the same session (probably because the proxy distinguishes `git-upload-pack` from `git-receive-pack`, or because fetch can degrade more gracefully than push).

## Why I'm not investigating

Per @venpopov 2026-05-20: first observation, possibly an Anthropic-side proxy change or a session quirk. Don't burn cycles until we see it again in a different session. This issue is the marker for that "next time."

## Workaround used

Direct push to `github.com` with the COO PAT (token var inlined into the URL). Then `git branch --unset-upstream` to avoid leaving the credential-bearing URL in `.git/config`. Caveat: inlining the variable into the command line means the expanded token appears in any shell-trace output — a credential-leak vector that the documented [`git-push-with-fallback.sh`](https://github.com/vade-app/vade-runtime/blob/main/scripts/git-push-with-fallback.sh) was designed to avoid; that fallback also failed in this session (companion issue below).

## If this recurs

Pick up the diagnostic at:
- the proxy log (if accessible) for the `WWW-Authenticate` round-trip
- whether `coo-bootstrap.sh` is supposed to install a credential helper (it doesn't currently — `git config credential.helper` is empty after a clean boot)
- whether the proxy's password expectation is documented anywhere in the Anthropic-side container setup

Sibling: companion issue below for silent-fail of `git-push-with-fallback.sh`.

[...]

PR #311, opened 2026-05-22
fix(git-push): surface rc=1 errors; ensure fallback fires on proxy rejection (closes #280)

### LINKED ISSUE #280: git-push-with-fallback.sh exits rc=1 silently under bootstrap-trace harness (observed-twice)
## Symptom

Single-session observation 2026-05-20 (run-2026-05-20T181315, cloud Claude Code container): [`vade-runtime/scripts/git-push-with-fallback.sh`](https://github.com/vade-app/vade-runtime/blob/main/scripts/git-push-with-fallback.sh) — the documented fallback for exactly the proxy failure described in [#279](https://github.com/vade-app/vade-runtime/issues/279) — also exits rc=1 with **no stdout, no stderr**.

This matters because the script's whole purpose is to be the loud, debuggable safety net under the proxy. Silent rc=1 from the safety net is the kind of failure that makes the next session conclude "git is broken" rather than "the fallback has a bug."

## Diagnostic

`bash -x scripts/git-push-with-fallback.sh -u origin <branch>` produced 20 lines of output — all bootstrap-trace harness preamble:

```
+ [[ 1 == \1 ]]
+ case "${0##*/}" in
+ [[ -n '' ]]
+ _VADE_BOOTSTRAP_TRACE_PROC_DONE=1
+ _VTRACE_ROOT=/root/.vade/traces
... (trace-run-id setup) ...
+ [[ ! -f /root/.vade/traces/<run-id>/meta.json ]]
+ exec
```

That's the preamble injected by [#273](https://github.com/vade-app/vade-runtime/pull/273) (bootstrap trace harness). After the bare `+ exec`, the trace shows nothing — neither the script's own statements nor any error. Both stdout and stderr files for the script were empty.

## Hypothesis (unverified)

The bootstrap-trace harness installs a wrapper that uses `exec` for redirection setup; in this script's context that interaction may be eating subsequent output (e.g. `set -e` exiting on an unhandled condition before the script writes its diagnostic). Pure hypothesis.

## Why I'm not investigating

Same as the sibling — single observation, may be a session-specific quirk of how the trace harness interacts with this script today. Per @venpopov 2026-05-20: don't burn cycles on transient-vs-real triage until we see it again.

## If this recurs

Start at:
- whether the bootstrap-trace harness ([#273](https://github.com/vade-app/vade-runtime/pull/273) / `coo-bootstrap-trace.sh`) is wrapping/intercepting scripts via `PROMPT_COMMAND`, `BASH_ENV`, or similar — and whether `git-push-with-fallback.sh` specifically interacts badly with that
- adding a trace-friendly `echo` at the script's top to confirm it's even reaching its body
- running the script with `env -i` to bypass the harness env and see if the body executes

Sibling: [#279](https://github.com/vade-app/vade-runtime/issues/279) (proxy 401 push failure).

https://claude.ai/code/session_01DWi5vs4hDPSYkxxRdhT9U8

---

## Closure gate

Closes when: **Track 6 wrapper-loudness lint** of the [secrets-management implementation](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/implementation-kickoff.md) lands. Class C silent-transport per the audit's [failure axis](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/reports/failure.md). Member of the git-push family (vrt#279, #280, #281, vade-coo-memory#898) grouped under vade-coo-memory#871. Sub-issue of vade-coo-memory#871; Project Status=Blocked.

PR #311 (closed 2026-05-23) attempted a per-symptom patch; per the triage, the wrapper-loudness lint (not a per-wrapper fix) is the designated closure vehicle.

_Per [triage row](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/triage.md). Enacted via vade-coo-memory#932._

vade-coo:
## Summary

Closes #280. The `git-push-with-fallback.sh` wrapper was observed twice exiting rc=1 with **zero stdout, zero stderr** — once in `run-2026-05-20T181315` under the bootstrap-trace harness ([#280 OP](https://github.com/vade-app/vade-runtime/issues/280#issue-)), then again in `run-2026-05-21T024648` while pushing a workflow-file commit ([#280 comment 2](https://github.com/vade-app/vade-runtime/issues/280#issuecomment-)). In both, the safety-net script became indistinguishable from "git is broken" to the next session.

The triggering case in observation 2 was the proxy emitting `refusing to allow an OAuth App ... workflow scope`, which **does** match `PROXY_FAILURE_PATTERNS` — so the fallback condition was satisfied. The reporter could not confirm whether (a) the rescue never fired, or (b) the rescue's output was also swallowed; from the user's seat both look identical.

## Root cause (best-supported reading)

stderr from the wrapper's `git push ... 2>&1 | tee/sed` invocations is routed through pipelines whose `sed`/`tee` ends can themselves be killed by SIGPIPE, closed fds, or harness-installed file-descriptor wrappers (the bootstrap-trace harness uses `exec {fd}>>...` to set up `BASH_XTRACEFD`). When stderr disappears from the user's view, the wrapper's existing `log_err` lines disappear with it, and the wrapper looks silent. The fallback push had no separate capture file at all (line was `git push ... 2>&1 | sed -E "$PAT_REDACT_SED"` with no tee), so even post-hoc inspection had nothing to read.

## Fix

Two changes, both in `scripts/git-push-with-fallback.sh`:

1. **Durable forensic log.** Every state transition (entry, primary push rc + byte-count, marker-match decision, fallback push rc, every error path) writes to `~/.vade/git-push-fallback.log` (overridable via `VADE_GIT_PUSH_FALLBACK_LOG`). When stderr is swallowed by an outer harness, the log preserves the boot-time history. Bounded retention (500 lines).
2. **Fallback push also tees.** The fallback pipeline gains a `| tee \"$ftmp\"` sink so the PAT-redacted output of a failed fallback lands in the durable log. Previously, double-fail cases (proxy + direct-URL both refused) left **no forensic trail at all**.

Additional hardening:
- `PIPESTATUS[0]:-1` default in case the shell handed back an empty value (paranoia, not a confirmed bug — see "remaining open questions").
- Every `return` path is now preceded by a `log_both` call (was previously `log_err`-only, easy to grep for). Naming `log_both` makes the durability contract explicit at the call site.
- Wrapper comment block documents the silent-failure hardening alongside the existing credential-leak hardening note.

## Verification

`scripts/ci/test-git-push-fallback.sh` is extended with two new test groups (13 new asserts, 31 total):

- **workflow-scope rejection (#280 observation 2).** Reproduces the proxy emitting the exact `refusing to allow an OAuth App ... workflow scope` string. Asserts: wrapper exits 0 after fallback rescue; stderr surfaces both the workflow-scope marker and the `retrying via https://vade-coo:***@github.com` message; the durable log records wrapper start, marker-match, and fallback rc=0.
- **double-fail forensic trail.** Both pushes fail. Asserts: wrapper exits 1; durable log dumps both pushes' raw output between `--- primary push ---` / `--- fallback push ---` fences; log captures `HTTP 403` and `fallback also rejected` content; log contains zero PAT material.

```
$ bash scripts/ci/test-git-push-fallback.sh
...
Results: 31 pass, 0 fail

$ bash scripts/ci/test-git-shim.sh
...
=== results: 13 passed, 0 failed ===
```

## Remaining open questions

- **The exact mechanism behind the swallowed stderr.** The hypothesis above (`BASH_XTRACEFD` / pipeline SIGPIPE / harness `exec` fd manipulation) is plausible but I could not reproduce the silent-stderr case in isolation — the workflow-scope rejection reproducer surfaces stderr cleanly. The fix is defensive (write to disk on every transition) rather than diagnostic of the exact stderr-disappearance route. If a future occurrence still surfaces silent stderr at the terminal, the `~/.vade/git-push-fallback.log` will at least say "wrapper start" + the rc + the captured push output, which is a 100% diagnostic improvement over the current "no output, exit 1" symptom.
- **Whether the bootstrap-trace harness in #273 ever needs to be told about this wrapper specifically.** If trace-mode output ends up routed via `BASH_XTRACEFD` and never reaches the user, that's an upstream harness concern (#273), not a wrapper concern. Filed as a "to consider" if #280 recurs after this fix.

## Test plan

- [x] `bash scripts/ci/test-git-push-fallback.sh` — 31/31 pass.
- [x] `bash scripts/ci/test-git-shim.sh` — 13/13 pass (no regressions).
- [x] `bash -n scripts/git-push-with-fallback.sh` — clean syntax.
- [x] This branch's own push surfaced `wrapper start` and `primary push: rc=0` lines, confirming the durable log + transition tracing fire on the happy path.
- [ ] Post-merge: on the next workflow-file push attempt, confirm `~/.vade/git-push-fallback.log` shows the wrapper's state transitions.

https://claude.ai/code/session_011dTuSjsfYFazcEecjpsHKR

### DIFF
--- a/scripts/ci/test-git-push-fallback.sh
+++ b/scripts/ci/test-git-push-fallback.sh
@@ -191,6 +191,175 @@ assert_match "fallback push args contain direct vade-coo URL" 'vade-coo:test_pat
 assert_no_match ".git/config contains no PAT" 'test_pat_DO_NOT_LEAK_2222' \
   "$(cat "$REPO/.git/config")"
 
+echo "== silent-failure hardening: workflow-scope rejection (#280) =="
+
+# Repro the #280 case directly: proxy rejects with the workflow-scope
+# OAuth message that should trip PROXY_FAILURE_PATTERNS=workflow.*scope.
+# The fix has to (a) match the marker, (b) fire the fallback, (c) leave
+# a diagnostic trail when the fallback itself fails so the next session
+# isn't stuck with rc=1 and zero output.
+
+WORKFLOW_WORK="$(mktemp -d)"
+WORKFLOW_LOG="$WORKFLOW_WORK/git-mock.log"
+WORKFLOW_STATE="$WORKFLOW_WORK/state"
+WORKFLOW_WRAPPER_LOG="$WORKFLOW_WORK/wrapper.log"
+
+cat > "$WORKFLOW_WORK/git" <<MOCKEOF
+#!/usr/bin/env bash
+case "\$1" in
+  push)
+    {
+      printf 'MOCK PUSH ARGS:'
+      for a in "\$@"; do printf ' [%s]' "\$a"; done
+      printf '\n'
+    } >> "$WORKFLOW_LOG"
+    if [ -f "$WORKFLOW_STATE" ]; then
+      # Fallback push: succeed.
+      exit 0
+    fi
+    : > "$WORKFLOW_STATE"
+    # Primary push: emit the canonical workflow-scope rejection.
+    cat >&2 <<'ERR'
+remote: refusing to allow an OAuth App to create or update workflow \`.github/workflows/foo.yml\` without \`workflow\` scope
+To http://127.0.0.1:35033/git/vade-app/vade-runtime
+ ! [remote rejected] claude/foo -> claude/foo (refusing to allow an OAuth App to create or update workflow without workflow scope)
+error: failed to push some refs to 'http://127.0.0.1:35033/git/vade-app/vade-runtime'
+ERR
+    exit 1
+    ;;
+  *)
+    exec $REAL_GIT "\$@"
+    ;;
+esac
+MOCKEOF
+chmod +x "$WORKFLOW_WORK/git"
+
+WORKFLOW_REPO="$WORKFLOW_WORK/repo"
+mkdir -p "$WORKFLOW_REPO"
+(
+  cd "$WORKFLOW_REPO"
+  $REAL_GIT init -q -b main >/dev/null 2>&1
+  $REAL_GIT -c user.email=t@t -c user.name=t commit --allow-empty -q -m init
+  $REAL_GIT checkout -q -b claude/workflow-test
+  $REAL_GIT remote add origin "http://local_proxy@127.0.0.1:8080/git/vade-app/vade-runtime"
+)
+
+WORKFLOW_STUB_HOME="$WORKFLOW_WORK/stub-home"
+mkdir -p "$WORKFLOW_STUB_HOME"
+
+(
+  cd "$WORKFLOW_REPO"
+  PATH="$WORKFLOW_WORK:$PATH"
+  HOME="$WORKFLOW_STUB_HOME" \
+  GITHUB_MCP_PAT="test_pat_workflow_280" \
+  VADE_GIT_PUSH_FALLBACK_LOG="$WORKFLOW_WRAPPER_LOG" \
+    bash "$WRAPPER" -u origin claude/workflow-test
+) > "$WORKFLOW_WORK/wrapper.out" 2>&1
+workflow_rc=$?
+
+workflow_out="$(cat "$WORKFLOW_WORK/wrapper.out")"
+
+assert_eq "workflow-scope: wrapper exits 0 after fallback rescue" "0" "$workflow_rc"
+assert_match "workflow-scope: stderr surfaced 'retrying via' message" \
+  'retrying via https://vade-coo:\*\*\*@github\.com' "$workflow_out"
+assert_match "workflow-scope: stderr surfaced the workflow-scope marker" \
+  'workflow.*scope' "$workflow_out"
+
+# Mock saw two pushes (primary proxy attempt + direct-URL fallback).
+workflow_push_count="$(grep -c '^MOCK PUSH ARGS:' "$WORKFLOW_LOG" 2>/dev/null || echo 0)"
+assert_eq "workflow-scope: mock saw two push attempts" "2" "$workflow_push_count"
+
+# Durable log was written.
+assert_match "workflow-scope: durable log records 'wrapper start'" \
+  'wrapper start: args=' "$(cat "$WORKFLOW_WRAPPER_LOG" 2>/dev/null || true)"
+assert_match "workflow-scope: durable log records 'proxy-failure marker matched'" \
+  'proxy-failure marker matched' "$(cat "$WORKFLOW_WRAPPER_LOG" 2>/dev/null || true)"
+assert_match "workflow-scope: durable log records fallback rc=0" \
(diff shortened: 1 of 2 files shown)


vade-coo:
Force-pushed: stripped `scripts/git-push-coo.sh` per your #310 close. PR now shows only the loudness fix on the existing `git-push-with-fallback.sh` + its test file (2-file diff, 280+/16-).

Reasoning (research-agent analysis on the secrets epic):

- The loudness fix is **Class-C wrapper-loudness work** — exactly what failure-axis report §C calls for ("loudness contract for every wrapper"). No entanglement with the secrets architecture; orthogonal to the schema, hooks, settings.json indirection, sub-agent env layers.
- The `git-push-coo.sh` wrapper that was carried in here belongs to **#310's scope**, which you correctly rejected. Same anti-pattern applied to merging it under a different number.
- Subtle finding worth flagging: the SOP at `coo/operations/secrets/README.md` §1 line 33 **already cites** `git-push-coo + GIT_ASKPASS shim` as the enforced defense (referencing vade-runtime#281). The plan **assumes** this wrapper exists without **authoring** it under any of the six tracks. Track 6's close-list expects #281 to land or move to follow-up. So the wrapper *is* planned-but-orphan — your #310 close note was structurally correct, and the authoring of `git-push-coo.sh` (along with the SOP citation that already references it) should be a proper Track-6 deliverable, coordinated with the SOP edits, not a one-off PR.

Bootstrap regression should re-run on the rebased branch shortly.

https://claude.ai/code/session_011dTuSjsfYFazcEecjpsHKR

vade-coo:
Closing as the same anti-pattern shape Ven called out on PR #310: per the [triage row](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/triage.md#related-not-under-871) for #280, the issue is held until Track 6 (wrapper-loudness lint) of the [secrets-management implementation](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/implementation-kickoff.md) lands. The lint — not a per-symptom patch — is the designated closure gate.

The intent here isn't to dismiss the fix — the rc=1 surfacing logic may well be correct — but the right vehicle is Track 6's wrapper-loudness pass, which audits *all* shell wrappers for silent-error patterns and lands a class-level lint preventing future regressions across #279, #280, #281, and vade-coo-memory#898.

Same trigger sequence as #310: this PR was opened immediately after the audit PR vade-coo-memory#905 landed but before the [triage](https://github.com/vade-app/vade-coo-memory/blob/main/coo/audits/2026-05-21-secrets/triage.md) was enacted, so the cross-reference wasn't surfaced. The enactment is now in progress at vade-coo-memory#932 (Briefing 034 follow-through).

Refs: [Briefing 034](https://github.com/vade-app/vade-coo-memory/blob/main/coo/briefings/034-secrets-audit-triage-followup.md) · Phase-A enactment plan vade-coo-memory#932.

PR #351, opened 2026-05-30
Aggregate output-styles/ subdir so outputStyle:coo loads on boot

vade-coo:
## What this fixes

On a fresh cloud boot the `coo` output style was inert — the system prompt came up as the stock default (git-workflow blocks present, COO override block absent). This was the RED post-merge confirmation of #348.

## Root cause

`outputStyle: coo` resolves `coo.md` from Claude Code's output-style search path (`<…>/.claude/output-styles/`). The boot aggregator unioned only `commands/agents/skills/hooks` into the effective `.claude/` — never `output-styles/`. So `coo-harness/.claude/output-styles/coo.md` was never symlinked in, the setting pointed at a missing file, and Claude Code fell back to the default prompt. Confirmed on this container: `~/.claude/output-styles/` and `/root/.claude/output-styles/` did not exist; `coo.md` existed only in the un-aggregated source.

## The fix

Add `output-styles` to the subdir loop in both projection functions in `scripts/lib/common.sh`:

- `sync_claude_config` (local/mac path)
- `aggregate_workspace_claude_config` (cloud path)

Two-line change. `aggregator.yml` governs the repo list, not subdirs, and is untouched.

## Verification

Called both real functions against a scratch `dst_root`:

- **Cloud path:** `output-styles/coo.md` projects as a symlink → `coo-harness/.claude/output-styles/coo.md`. ✅
- **Local path:** same, as a dir-level symlink. ✅

`integrity-check.sh` does not assert the subdir set, so no invariant is affected; the bootstrap-regression CI exercises the cloud path end-to-end.

## Scope (per handoff)

Aggregator-only. Deliberately **not** touched here:

- The descriptive enumeration comments at `common.sh:383` / `:499` and `coo-memory/operations/BOOT_TOPOLOGY.md` are now slightly stale — follow-up doc sync.
- The `CLAUDE.md` override-paragraph prune **stays blocked** until a fresh post-fix boot proves the `coo` block actually loads in-prompt.

## Boot-impacting

Verification can only happen on a fresh container boot. A self-contained handoff prompt (pre-merge gating + post-merge confirmation) follows as a standalone comment.

Closes: n/a

Follow-up to #348 — resolves the RED post-merge confirmation of that PR; no standalone issue tracks it.

https://claude.ai/code/session_01FQNoUhHfi3Xm9EDS4G2ApA

PR #354, opened 2026-05-30
fix(boot): force-disable git instructions via env var on cloud

vade-coo:
## What this does

Adds `CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS=1` to the static `env` block in `.claude/settings.json`.

## Why

On the Claude Code **web/cloud** surface, `includeGitInstructions: false` is **not honored** — the `## Git Operations` block (commit/PR workflow + git-status snapshot) survives in the system prompt despite the setting. Verified first-hand on a 2026-05-30 fresh boot — the same boot that ran the post-merge confirmation of #351 (output-style aggregation).

Per the [settings docs](https://code.claude.com/docs/en/settings.md), `CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS` *"takes precedence over this setting when set."* It's the lever that bites on cloud where the settings.json key doesn't.

## Scope — what this does and does NOT remove

This env var (and `includeGitInstructions`) govern **only** the `## Git Operations` block. The other two git-shaped blocks in the cloud system prompt are **separate web-platform injections** this lever does not reach:

- `## Git Development Branch Requirements` — actually *correct* for cloud (the gitproxy gates push to the session branch); we would not want to suppress it.
- `## GitHub Integration` ("no `gh` CLI, use MCP") — the load-bearing contradiction with our canonical `gh` write path. No documented lever; the output-style override (`coo.md` standing-override #1) stays load-bearing for this regardless.

So this is a partial cleanup, not a fix for the `gh`-vs-MCP contradiction — and the CLAUDE.md override-paragraph prune **stays blocked**.

## Governance

Static committed env var — same class as `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS`, **not** a runtime-injected secret threaded through `_write_claude_settings_env`'s positional args. So this is *not* the "coordinated change / explicit approval" surface named in `CLAUDE.md`. Propagation verified against `_sync_claude_settings` (`scripts/lib/common.sh:588` — `merged.env = {...src.env, ...dstEnv}`: source keys land, existing secrets overlay, nothing deleted).

## Verification

Boot-impacting — effect is only observable on a fresh container boot. A handoff prompt (pre-merge gating + post-merge confirmation) follows as a standalone comment.

Follow-up to #351; resolves the `## Git Operations` half of that PR's open question.

Closes: n/a

https://claude.ai/code/session_0163ZbtN5Bsm77vHC6RRbUdY

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -118,7 +118,8 @@
   },
   "env": {
     "CLAUDE_STREAM_IDLE_TIMEOUT_MS": "600000",
-    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS": "1"
+    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS": "1",
+    "CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS": "1"
   },
   "permissions": {
     "allow": [


vade-coo:
## Handoff prompt — post-merge verification (boot-impacting)

Env vars are read at process launch, so this change's effect can only be confirmed on a **fresh boot after it lands in `main`** (a new container clones `main`; this branch's prompt is already built and cannot be re-derived mid-session). CI is parse/merge-path only; the real proof is post-merge.

### Pre-merge gating
- [ ] `bootstrap-regression` CI green — touching `.claude/` triggers it; it runs `cloud-setup.sh` → `session-start-sync.sh` → `integrity-check.sh`, exercising `_sync_claude_settings` (the env-propagation path) and asserting the integrity report.
- [ ] Diff is exactly the one added line in the `.claude/settings.json` `env` block — nothing else.

### Post-merge confirmation (run on a fresh boot)
1. **Var present in effective settings** — expect `1` on both roots, and set in the live env:
   ```sh
   jq -r '.env.CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS' /root/.claude/settings.json /home/user/.claude/settings.json
   printf 'live env: [%s]\n' "${CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS:-<UNSET>}"
   ```
2. **`## Git Operations` gone** — confirm the git push/fetch retry block ("Always use git push -u origin …") is **absent** from the system prompt.
3. **What SHOULD remain** (separate web-platform injections, *not* governed by this var — their presence is expected, not a regression):
   - `## Git Development Branch Requirements` (names the session `claude/*` branch; the gitproxy actually enforces it)
   - `## GitHub Integration` ("no `gh` CLI, use MCP")

### Decision tree
- **`## Git Operations` gone** → the env var works on cloud; static settings.json `env` is confirmed as the cloud lever. Then update `MEMO-2026-05-30-g5zq`'s "removed by `includeGitInstructions: false`" line to name `CLAUDE_CODE_DISABLE_GIT_INSTRUCTIONS` as the cloud mechanism (settings key bites on local/CLI only).
- **`## Git Operations` still present** but the var is confirmed in env → `"1"` may be the wrong token; retry with `"true"`. If `"true"` also fails → the block is injected by a path even this env var doesn't reach (like the other two), so revert this change as inert and escalate.

Either outcome: the CLAUDE.md override-paragraph **prune stays blocked** — the `## GitHub Integration` contradiction persists regardless of this var, so the output-style override remains load-bearing.

https://claude.ai/code/session_0163ZbtN5Bsm77vHC6RRbUdY

PR #383, opened 2026-05-30
project-board-digest: server-side status filter via ProjectV2.items.query (closes #350, #352)

### LINKED ISSUE #350: project-board-digest: 910-pt full-board fetch per boot drains shared GraphQL budget
## Summary

On 2026-05-30 the `vade-coo` PAT's GraphQL budget (5,000 points/hr, **per-user**, shared across every session and the github MCP server) hit zero and stayed there for ~50 minutes. The first interactive session of the day booted into an empty bucket and could not read the project board until the hourly reset.

Root cause: `scripts/lifecycle/project-board-digest.sh` runs as a SessionStart hook and makes a **910-point** full-board GraphQL fetch on *every* boot, and the image build warm-booted the container ~15× in three minutes. Roughly 5–6 of those fetches alone exhaust the budget.

## Measured cost (today, immediately after reset)

- One `gh project item-list 1 --owner coo-labs --limit 1000` = **910 GraphQL points** (measured: `graphql_used` 0 → 910 on a single call).
- Board size: **819 items** — Done 606, Todo 144, In Progress 37, Blocked 17, Paused 15.
- The digest only emits the **37 In Progress** items, so we fetch **819 to use 37 (~22× over-fetch)**.
- **5,000 ÷ 910 ≈ 5.5 fetches exhausts the whole hourly budget.**
- The script comment and CLAUDE.md both say "~370 items" — **stale by 2.2×**. The board has grown ~450 items in the nine days since the digest landed (#813), largely coo-logs session-log issues being tracked as board items.

## Evidence (this container's logs)

- `coo-bootstrap.log`: ~10 "cached PAT validated" warm-boots between 08:38:17 and 08:41:02, ~15–22s apart.
- `boot.log`: ≥4 distinct session IDs (`98fd1627`, `26472446`, `64128271`, `270f9ee2`, …) each running `project-board-digest` start/end in the 08:40–08:41 window.
- `transcript-export-logs/`: 10+ exports, distinct PIDs, starting 08:38:40Z.
- The interactive session booted 09:12; its digest 429'd → "project item-list failed; skipping digest".
- `rate_limit` at 09:16: graphql **5000/5000** used, core **2**/5000 — exclusively GraphQL.

## Why the over-fetch is forced

Projects v2 GraphQL has no server-side field filter; the `items` connection returns the whole board and the script filters `In Progress` client-side with jq. The full-board cost is therefore unavoidable *per call* — which is exactly why it must be paid **once and cached**, never multiplied by boot count.

## Root cause — two compounding defects

1. **Network-touching SessionStart hooks aren't gated against build-time warm-boots.** `project-board-digest.sh` and `discussions-digest.sh` hit the live, rate-limited API on every boot. Build warm-boots multiply the spend by the boot count and burn the budget the real interactive session needs. The warm-boot signal already exists (the bootstrap logs "cached PAT validated").
2. **The fetch costs 910 points because the board is 819 items, 606 of them Done.** Completed items are never archived off the board, so every digest pages through 606 finished items to surface 37 active ones.

## Proposed fixes (priority order)

1. **Gate the live-API digests to no-op during image-build warm-boots** (env flag set by the build harness, or reuse the existing warm-boot signal). Highest leverage — removes the multiplier.
2. **Cache the digest**: one scheduled job (~every 15 min) performs the unavoidable fetch and writes an artifact; boots read the file. Turns N fetches per boot-burst into ≤1 per 15 min.
3. **Archive Done items off the board** (606 → board shrinks ~74%, fetch cost drops proportionally), and stop tracking coo-logs session-log issues as board items.
4. **Lean query**: replace `gh project item-list` with a hand-rolled `gh api graphql` requesting only the four fields the digest needs, cutting per-fetch points.
5. **Split the bucket**: route read-heavy automation onto the `vade-coo-app` installation token, which has its own per-installation rate limit separate from the per-user PAT.

## Secondary anomaly — warrants its own investigation

The image build booting Claude **~15 times in three minutes** (and exporting ~15 transcripts) is itself unexpected; a normal build should not warm-boot the agent that many times. At 910 points/fetch only ~5 boots are needed to exhaust GraphQL, so the boot count both caused this incident and points at a separate build-harness / OS-image-pipeline defect. That driver likely lives outside coo-harness — recommend a dedicated issue once it is identified.

## Refs

- Script: `scripts/lifecycle/project-board-digest.sh`
- Digest introduced: coo-labs/coo-memory#813
- Stale "~370 items" claim also in `coo-memory/CLAUDE.md` §9 and the script comment.

https://claude.ai/code/session_01FQNoUhHfi3Xm9EDS4G2ApA

### LINKED ISSUE #352: Boot digests fire unconditional GraphQL on every startup — amplifier on shared vade-coo quota
## Summary

Two SessionStart digests — `project-board-digest.sh` and `discussions-digest.sh` — each fire an **unconditional GraphQL fetch on every `matcher: startup` boot**, with no dedup, no cache, and no cooldown. `project-board-digest.sh` pulls all ~370 project items (`gh project item-list 1 --limit 1000`); `discussions-digest.sh` runs its own GraphQL query. So **N boots of a container = N full fetches**, and every container authenticates as the same `vade-coo` identity against one shared 5000-point/hr GraphQL quota.

A normal idle-resume is cheap (resumes don't re-run the `startup` chain). But any container that boot-loops or is rapidly re-spawned turns each boot into two expensive GraphQL queries, and the cost lands on the **shared** quota — degrading every other concurrent COO session.

## Evidence

Cross-container symptom observed 2026-05-30 (reported by a parallel session, reading a different container's logs):

- `~/.vade/boot.log`: four distinct session boots ~15s apart in a burst — `98fd1627…` 08:40:18, `26472446…` 08:40:32, `64128271…` 08:40:46, `270f9ee2…` 08:41:01 — **each** running `project-board-digest → phase:end ok` (full 370-item fetch) plus `discussions-digest`.
- `~/.vade/coo-bootstrap.log`: same cadence independently — ~10 "cached PAT validated" boots between 08:38:17 and 08:41:02, one every ~15–22s.
- `transcript-export-logs/`: ten-plus export files from 08:38:40Z, each a separate PID, ~15–22s apart.

Blast radius confirmed from an unrelated healthy container ~1.5h later: GraphQL quota at **2027/5000 remaining (2973 used)**, and a routine `gh pr list` in that session was rate-limited as collateral. The burst session IDs are absent from the healthy container's filesystem (`~/.vade` is container-local on `/dev/vda`, no shared mount), confirming the amplifier is per-container but the **quota is shared**.

## Root cause vs. trigger

- **Root cause (this issue):** the digests are unguarded against repeated boots. This is a standing amplifier independent of *why* a container boots repeatedly.
- **Trigger (out of scope here):** what caused the 15s re-spawn cadence in that container is a separate question — candidates include a repeatedly-spawned fresh session for the coo-labs/coo-harness#348 post-merge verification, or a boot-crash loop. Not diagnosable from inside a different container; tracked separately if it recurs.

## Proposed fix

Add a short-TTL cooldown to both digests: a container-local marker (e.g. `~/.vade/agent-state/<digest>.last`) recording the last successful fetch. On boot, if the marker is younger than TTL (e.g. 5–10 min), **reuse the last digest output** instead of re-querying. Caps per-container, per-window GraphQL cost regardless of boot frequency, and turns a boot loop from a shared-quota incident back into a cheap non-event.

Both scripts already no-op gracefully on missing token / failed fetch — the cooldown is the same defensive posture extended to "fetched too recently."

## Notes

- Boot-script change → boot-impact PR convention applies (handoff prompt + fresh-boot verification).
- Surfaced during the coo-labs/coo-harness#348 post-merge verification session; not caused by #348 (an unrelated healthy container carries identical `settings.json` and boots fine).

https://claude.ai/code/session_01MAiZACKi5jdp8UbcEw7QqH

vade-coo:
## Summary

Replaces `gh project item-list 1 --limit 1000` (full-board fetch, ~910 GraphQL pts, all 819 items, jq-filtered client-side for 18 In Progress) with a direct GraphQL call using the **undocumented `query` argument** on `ProjectV2.items`:

```graphql
items(first: 100, query: "status:\"In Progress\"") { ... }
```

Single call, ~1 GraphQL pt, returns only matching items in one page. The `query` arg matches the project UI filter DSL, isn't in the published API reference, but is present in the live schema and verified working against project #1 (introspection + 3 independent live calls 2026-05-30).

Output is byte-equivalent to the prior jq — same group-by-milestone, same `(no milestone)` placement, same `repo#N — title` rows.

Removes the temporary disable guard added in #360. The quota concern that motivated it is now ~900× smaller per fetch; cooldown / cache complexity is no longer load-bearing and was dropped from this PR.

## Closing keywords

Closes #350
Closes #352

## Cross-repo references

- Follow-up needed in `coo-labs/coo-memory`: revert CLAUDE.md §9 + `identity/episodic_memory.md` "Active work surface" text introduced alongside #360 — the digest is back online with the cheap fetch. File after fresh-boot verification.

## Verification

Local-tested in this session against live data:

| Run | Duration | GraphQL Δ | Output |
|---|---|---|---|
| Old `gh project item-list --limit 1000` | 11.8s | ~405 pts | 15 items, 3 milestones |
| New `gh api graphql query:"status:..."` | 2.8s | ~2 pts | 18 items, 3 milestones (identical shape) |

`bash -n` clean. Coverage cap `first: 100` flagged in jq via `[+N more — raise first:N]` if In Progress ever exceeds the page.

## Why "undocumented but trust it"

This is the second Projects v2 API surface in two weeks where the docs lag the schema — the `issueFieldValues` connection used in `operations/issue-fields-and-types.md` was found the same way (introspection-first, docs-second). The argument has zero changelog mention but is present in the GA schema. Trusting introspection is the pattern, per MEMO worth noting as a standing reminder.

https://claude.ai/code/session_01Mzr2s8j6wRE3EmWRY7aePQ

### DIFF
--- a/scripts/lifecycle/project-board-digest.sh
+++ b/scripts/lifecycle/project-board-digest.sh
@@ -20,11 +20,6 @@ trap '_rc=$?; boot_log_record project-board-digest end $([ $_rc -eq 0 ] && echo 
 
 wait_for_coo_bootstrap 60
 
-# TEMPORARILY DISABLED — GitHub project / API quota investigation in flight.
-# Remove this guard to restore the digest fetch.
-echo "Boot: VADE project digest — TEMPORARILY DISABLED (GitHub project fetch / API quota investigation in flight). Do not query the project board this session unless explicitly asked."
-exit 0
-
 TOKEN="${GH_TOKEN:-${GITHUB_TOKEN:-${GITHUB_MCP_PAT:-}}}"
 if [ -z "$TOKEN" ]; then
   log "GH_TOKEN unset; skipping project-board digest."
@@ -36,35 +31,68 @@ if ! check_cmd gh || ! check_cmd jq; then
   exit 0
 fi
 
-# Project 1 currently holds ~370 items; --limit must exceed total or the
-# fetch silently truncates. 1000 gives comfortable headroom.
-RESPONSE="$(GH_TOKEN="$TOKEN" gh project item-list 1 --owner coo-labs \
-  --format json --limit 1000 2>/dev/null || echo '')"
+# The 2026-05-30 quota incident: `gh project item-list 1 --limit 1000`
+# paged the whole 819-item board for ~910 GraphQL pts to surface ~37 In
+# Progress items, and a build-time warm-burst multiplied that until the
+# 5000-pt/hr quota drained.
+#
+# Fix: server-side filter via the undocumented `query` argument on
+# `ProjectV2.items` (confirmed via schema introspection 2026-05-30; not
+# in the published API reference but matches the project UI filter DSL).
+# Single call, ~1 GraphQL pt, returns only matching items in one page.
+#
+# Coverage cap: `first: 100` — alarms if In Progress ever exceeds that.
+RESPONSE="$(GH_TOKEN="$TOKEN" gh api graphql -f query='
+{
+  organization(login: "coo-labs") {
+    projectV2(number: 1) {
+      items(first: 100, query: "status:\"In Progress\"") {
+        totalCount
+        pageInfo { hasNextPage }
+        nodes {
+          content {
+            __typename
+            ... on Issue {
+              number title
+              repository { name }
+              milestone { title }
+            }
+            ... on PullRequest {
+              number title
+              repository { name }
+              milestone { title }
+            }
+          }
+        }
+      }
+    }
+  }
+}' 2>/dev/null || echo '')"
 
 if [ -z "$RESPONSE" ]; then
-  log "project item-list failed; skipping digest."
+  log "project board GraphQL fetch failed; skipping digest."
   exit 0
 fi
 
-# Echo the formatted block. jq does all the work — group by milestone,
-# sort milestones alphabetically (with "(no milestone)" last), then
-# emit one section per milestone.
+# Format: group by issue milestone, sort milestones alphabetically with
+# "(no milestone)" last via the "~no-milestone" sort key.
 printf '%s' "$RESPONSE" | jq -r '
-  [.items[] | select(.status == "In Progress")] as $ip
+  (.data.organization.projectV2.items // {nodes: [], totalCount: 0, pageInfo: {hasNextPage: false}}) as $items
+  | $items.nodes as $ip
   | if ($ip | length) == 0 then
       "Boot: VADE project — no items In Progress."
     else
       ($ip
-        | group_by(.milestone.title // "~no-milestone")
-        | sort_by(.[0].milestone.title // "~no-milestone")
+        | group_by(.content.milestone.title // "~no-milestone")
+        | sort_by(.[0].content.milestone.title // "~no-milestone")
       ) as $groups
       | (
-          "Boot: VADE project — In Progress (\($ip | length) item\(if ($ip|length)==1 then "" else "s" end) across \($groups | length) milestone\(if ($groups|length)==1 then "" else "s" end)):"
+          "Boot: VADE project — In Progress (\($ip | length) item\(if ($ip|length)==1 then "" else "s" end) across \($groups | length) milestone\(if ($groups|length)==1 then "" else "s" end))\(if $items.pageInfo.hasNextPage then " [+\($items.totalCount - ($ip|length)) more — raise first:N]" else "" end):"
         ),
         ($groups[] | (
-          "  [\(.[0].milestone.title // "(no milestone)")]"
+          "  [\(.[0].content.milestone.title // "(no milestone)")]"
         ), (
-          .[] | "    \(.repository | sub("https://github.com/coo-labs/"; ""))#\(.content.number) — \(.title)"
+          .[] | "    \(.content.repository.name)#\(.content.number) — \(.content.title)"
         )),
         "URL: github.com/orgs/coo-labs/projects/1 (filter status:\"In Progress\")"
     end


vade-coo:
PR #383 fixes the 2026-05-30 GraphQL-quota drain by adding TTL cache to the two SessionStart digests and removes the `#360` safety disable. Verification needs both branch-side review and a fresh-container boot.

## Pre-merge gating

1. Diff review against the issue analyses:

   ```sh
   gh pr diff 383 --repo coo-labs/coo-harness
   ```

   Confirm: `digest_cache_{fresh,read,write}` helpers added to `scripts/lib/common.sh`; both digest scripts gated by `digest_cache_fresh` before any network call; project-board-digest `exit 0` guard removed; `tee` pattern preserves stdout while writing cache.

2. (Optional) Local cooldown test on the branch:

   ```sh
   gh pr checkout 383 --repo coo-labs/coo-harness
   rm -f "$HOME/.vade/agent-state/project-board-digest."{cache,last}
   rm -f "$HOME/.vade/agent-state/discussions-digest."{cache,last}
   BEFORE=$(gh api rate_limit --jq '.resources.graphql.used')
   bash scripts/lifecycle/project-board-digest.sh > /tmp/cold.txt
   bash scripts/lifecycle/project-board-digest.sh > /tmp/warm.txt
   AFTER=$(gh api rate_limit --jq '.resources.graphql.used')
   echo "GraphQL used: $((AFTER - BEFORE)) pts (expect ~400–910, NOT ~800–1820)"
   diff /tmp/cold.txt /tmp/warm.txt && echo "cache stable ✓"
   ```

   Expect: cold + warm together cost ≤ 1 fetch (cache hit on the second). If the delta exceeds ~910 pts, the cooldown isn't gating — do NOT merge.

## Post-merge confirmation

The cooldown's load-bearing path — gating against a *boot burst* of fresh containers — only fires on real SessionStart. Run on the next fresh container after merge:

1. **First fresh boot.** The boot system-reminder should show:

   ```
   Boot: VADE project — In Progress (N items across M milestones):
     [...]
   ```

   …NOT the `TEMPORARILY DISABLED` line. If it still shows DISABLED, the `#360` guard wasn't removed cleanly.

2. **Cache populated.** In the first session:

   ```sh
   ls -la "$HOME/.vade/agent-state/"{project-board-digest,discussions-digest}.{cache,last}
   ```

   Expect four files, all written within the last minute, all non-empty.

3. **Quota healthy.**

   ```sh
   gh api rate_limit --jq '.resources.graphql | "GraphQL: \(.used)/\(.limit), reset \(.reset | strftime("%H:%M UTC"))"'
   ```

   Expect `<1000/5000` after one fresh boot. If higher, something is mis-counting.

4. **Second fresh boot (within 30 min of the first).** Boot a second fresh container and check `~/.vade/boot.log` for `project-board-digest end ok`; agent-side `log` lines should include `cache hit (ttl=1800s); skipping fetch.` If the digest fetched again, the cache isn't being read across containers (expected — cache is per-container) OR the marker isn't honored.

   Note: cache is intentionally per-container (`~/.vade` is on `/dev/vda`). Cross-container amplification is a *separate* concern — see #350 fix #2 (scheduled job writing a shared artifact) and #5 (app-installation-token routing). Not in scope here.

## Follow-ups blocked by this PR

- **`coo-labs/coo-memory` revert PR**: CLAUDE.md §9 and `identity/episodic_memory.md` "Active work surface" both carry text saying the project-board digest is disabled. Once steps 1–3 above pass, file a small PR reverting that text so the substrate matches reality.
- The 15-boot warm-burst itself (separate from the amplifier this PR closes) is **#350 §"Secondary anomaly"** — out of scope; consider opening a dedicated issue if it recurs.

## Greeting

Greet Ven after confirmation in step 3 with a one-line "PR #383 verified on fresh boot, GraphQL quota healthy, follow-up cross-repo revert filed/queued." Don't greet before — the verification is the deliverable.

https://claude.ai/code/session_01Mzr2s8j6wRE3EmWRY7aePQ

vade-coo:
**Approach pivot (force-pushed cd89944).** The cooldown/cache approach in the prior commit was based on a wrong premise inherited from #350's analysis ("Projects v2 GraphQL has no server-side field filter"). Live schema introspection 2026-05-30 disproved that: `ProjectV2.items` accepts an undocumented `query: String` argument that takes the project UI filter DSL.

Server-side filter `status:"In Progress"` cuts per-fetch cost from ~910 GraphQL pts to ~1, in ~2.8s vs ~11.8s. The cooldown / cache complexity was dropped — the filter makes it unnecessary at any reasonable boot rate.

Verification simplified accordingly — no TTL behavior to test across container boots, just "does the digest fetch and emit the right items on cold boot." The original handoff comment above is superseded; the new verification is just:

1. Fresh container boot → boot system-reminder shows `Boot: VADE project — In Progress (N items across M milestones):`, NOT the DISABLED line.
2. `gh api rate_limit` after one boot shows GraphQL usage <50/5000.
3. File the `coo-memory` revert PR (CLAUDE.md §9 + episodic_memory.md).

https://claude.ai/code/session_01Mzr2s8j6wRE3EmWRY7aePQ

PR #433, opened 2026-06-04
secrets(track-4): schema-iterator + SA break-glass + #66 strip-on-failure (Phase 1)

vade-coo:
Track 4 Phase 1 of the secrets-management implementation (coo-labs/coo-memory#871): schema-iterator refactor of `fetch_coo_secrets`, SA-token break-glass, and coo-harness#66 strip-on-failure.

**Tracks:** Track 4 Phase 1 of coo-labs/coo-memory#871. Sibling to Tracks 1/2/3/5 shipped this session.

### (a) `coo-bootstrap.sh` schema-iterator refactor

Replaces the 190-line hardcoded `fetch_coo_secrets` block in `scripts/lib/common.sh` with a call to the new `scripts/boot/schema-fetch-secrets.py` helper. The helper reads `operations/secrets/schema.yaml` (via `VADE_COO_MEMORY_DIR`), iterates `status: active` credentials, and emits eval-safe `export VAR=VALUE` lines. Status discipline (dormant/retired/pending_* = skip, unknown = warn) is enforced in the helper. Multi-field credentials (`r2-transcripts`, `cloudflare-api-vade-coo`, `github-app-vade-coo-app`) handled via `MULTI_FIELD_MAP`. Fail-closed if schema is unreadable or PyYAML is missing.

### (b) SA-token break-glass

- **(b1)** `op whoami --format=json` identity check early in bootstrap. On first successful boot, writes the SA name to `~/.vade/.op-sa-identity`. Subsequent boots compare; identity shift → exit non-zero + inline recovery procedure citing SOP §3.6.
- **(b2)** Two-token-overlap: if `OP_SERVICE_ACCOUNT_TOKEN_NEW` is set (Ven rotating SA token), tries the new token first; on success, persists it and unsets `_NEW`. On failure, falls back to current token.
- **(b3)** Manual override path: `OP_SERVICE_ACCOUNT_TOKEN=<new> bash coo-bootstrap.sh` is preserved and documented in error output. A companion coo-memory PR will add the exact recovery commands to SOP §3.6 explicitly.

### (c) coo-harness#66 strip-on-failure

[...]

PR #434, opened 2026-06-04
secrets(track-4): patch op whoami JSON key list for op 2.31+ schema

vade-coo:
Phase 1 (#433) follow-up. The SA-identity self-discovery check it introduced silently no-ops on the production op CLI version; the break-glass safeguard it was meant to install is non-functional.

## The bug

`scripts/boot/coo-bootstrap.sh:299` calls `op whoami --format=json` and parses the result with a Python fallback list:

```python
for key in ['ServiceAccount', 'service_account', 'name', 'email', 'user_email']:
```

That list matches the bootstrap-regression CI mock at `scripts/ci/mocks/op` but not the live op 2.31+ schema:

```json
{"url":"...","URL":"...","user_uuid":"...","account_uuid":"...","user_type":"SERVICE_ACCOUNT","ServiceAccountType":"SERVICE_ACCOUNT"}
```

No key in the parser list exists in the live JSON. `_sa_name` is empty, the else branch fires:

```
1Password service account authenticated (identity field not found in JSON; skipping identity check)
```

`~/.vade/.op-sa-identity` is never written. The comparison logic that's supposed to detect SA-token-identity rotation never runs. CI passes (mock returns `ServiceAccount`); production silently skips.

## The fix

- Prepend `user_uuid`, `account_uuid` to the key list. `user_uuid` is the stable per-integration identifier — the right unit for SA-rotation detection. `account_uuid` as a secondary fallback if op ever renames the integration field.
- Update the schema comment in the parser to reflect the live op 2.31+ shape.
- Update `scripts/ci/mocks/op` so `--format=json` emits the live schema. Legacy `ServiceAccount` is kept in parallel so forward-compat across CLI versions stays exercised.

## Verification (locally, before push)

Cold-bootstrap on the production container:

1. Deleted `/root/.vade/.coo-bootstrap-done`.
2. Ran `CLAUDE_CODE_REMOTE=true bash scripts/boot/coo-bootstrap.sh` end-to-end. Completed clean (`coo-bootstrap: complete`).
3. `/root/.vade/.op-sa-identity` now exists, contents = the `user_uuid` returned by live `op whoami --format=json`.
4. Re-ran `integrity-check.sh`: 30/35 DEGRADED — exact same `F1,S2,S3,S7,S8` baseline as before. No new degradation introduced.

## Why this matters

Track 4 Phase 1's premise is that SA-token rotation can be detected and refused at boot. With the silent skip, any SA token whose `op whoami` succeeds will pass, regardless of whether it's the expected `vade-coo` SA or a different one. Phase 2 stacks more secrets-flow changes on top; fixing this before Phase 2 lands keeps Phase 2's blast radius narrower.

## Refs

- Umbrella: coo-labs/coo-memory#871
- Phase 1 PR: #433 (introduced the parser)
- SOP: [`operations/secrets/README.md`](https://github.com/coo-labs/coo-memory/blob/main/operations/secrets/README.md) §3.6 (recovery doc — referenced from the bootstrap when identity mismatch fires)

https://claude.ai/code/session_01ScN2pdkYKkm5YdKhxZWrRN

### DIFF
--- a/scripts/boot/coo-bootstrap.sh
+++ b/scripts/boot/coo-bootstrap.sh
@@ -301,12 +301,15 @@ _whoami_json=""
 if _whoami_json="$(op whoami --format=json 2>/dev/null)"; then
   # Extract service account identifier from JSON output. op whoami --format=json
   # returns different schemas across versions; try multiple field names.
+  # op 2.31+ live: {'url','URL','user_uuid','account_uuid','user_type','ServiceAccountType'}
+  # — `user_uuid` is the stable per-integration identifier and the right
+  # unit for SA-rotation detection. Older schemas exposing `ServiceAccount`
+  # / `name` are kept for forward-compat across CLI versions.
   _sa_name="$(printf '%s' "$_whoami_json" | python3 -c "
 import json, sys
 try:
     d = json.load(sys.stdin)
-    # op 2.x: {'ServiceAccount': 'name', 'URL': ..., 'UserUUID': ...}
-    for key in ['ServiceAccount', 'service_account', 'name', 'email', 'user_email']:
+    for key in ['user_uuid', 'account_uuid', 'ServiceAccount', 'service_account', 'name', 'email', 'user_email']:
         if key in d and d[key]:
             print(d[key]); break
 except Exception:
--- a/scripts/ci/mocks/op
+++ b/scripts/ci/mocks/op
@@ -26,11 +26,13 @@ case "${1:-}" in
     ;;
   whoami)
     # Track 4 Phase 1: support --format=json for SA identity check.
-    # Return the same ServiceAccount name in both text and JSON forms
-    # so the identity file is written and compared correctly in CI.
+    # Returns op 2.31+ live schema (url/URL/user_uuid/account_uuid/user_type/
+    # ServiceAccountType) so the CI mock matches the production CLI shape;
+    # `ServiceAccount` is kept in parallel so forward-compat across CLI
+    # versions stays exercised in CI.
     case "${2:-}" in
       --format=json|--format)
-        printf '{"ServiceAccount":"ci-mock-coo-bootstrap","URL":"https://ci.mock","UserUUID":"00000000-0000-0000-0000-000000000000"}\n'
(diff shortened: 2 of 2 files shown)


vade-coo:
## Handoff prompt — fresh-boot verification

This PR fixes a silent skip in Phase 1's SA-identity self-discovery. The fix was cold-verified by deleting the bootstrap marker and re-running `coo-bootstrap.sh` in-container, but the production verification surface is a true fresh-container cold boot. Run the following from the *next* session that boots against the post-merge `main`.

2. `identity/charter.md`
3. `identity/governance.md`
4. `identity/preferences.md`

These can all be issued in one parallel batch. Once the brake releases, the verification commands below will work.

### What to check (cap output to 8 lines per check)

1. **Boot digest expected-yellow only.** `summary.ok=false` with degraded set ⊆ `{S2,S3,S7,S8}` (SOP §3.4). F1 may still hold the two prior-session citation drift commits (`16cb6fb`, `21fa1f6`) — note but do not act on.
2. **\`/root/.vade/.op-sa-identity\` exists.** \`ls -la /root/.vade/.op-sa-identity\` returns a non-empty file. Content matches \`op whoami --format=json | jq -r .user_uuid\` — currently \`CJABPIXKYBCGJFX3SWSQF55FXI\` for the live \`vade-coo\` SA. If the content differs, the SA token in the container env is not the expected one — surface to Ven, do not proceed.
3. **Bootstrap log shows the success-branch log line.** \`grep -F "1Password service account authenticated" /root/.vade/coo-bootstrap.log\` returns a line like \`1Password service account authenticated: CJABPIXKYBCGJFX3SWSQF55FXI\` — NOT \`(identity field not found in JSON; skipping identity check)\`. The latter wording is the bug shape and indicates the patch did not land.
4. **\`op whoami\` still works.** Plain \`op whoami\` exits 0 and reports \`User Type: SERVICE_ACCOUNT\`.

### Rollback shape (if any check fails)

\`git revert <merge-commit>\` on \`coo-harness/main\` returns the parser to the Phase 1 silent-skip state. That posture passed prior boots; rollback is safe. After revert, Phase 2 dispatch is blocked until a forward fix lands.

### After verification

Reply to coo-memory#871 with a one-line verification receipt: \`Phase 1 SA-identity self-discovery verified on fresh boot — .op-sa-identity = <uuid>\`. Then proceed to Track 4 Phase 2 dispatch per the kickoff prompt at coo-labs/coo-memory#871 (comment 4621355098).

### Refs

- This PR introduces a one-line fix to \`scripts/boot/coo-bootstrap.sh:309\` (key-list prepend) plus mock alignment at \`scripts/ci/mocks/op\`.
- Phase 1 PR: #433.
- Umbrella: coo-labs/coo-memory#871.

---

*Edited 2026-06-04: prior version asserted the identity-stack Read could be waived. That was wrong — the boot brake (MEMO-2026-06-04-nzxf) mandates it. Substantive verification steps unchanged.*

PR #443, opened 2026-06-05
op-coo-wrap: transparent SA-token rate-limit fallback

vade-coo:
## What

Path-shadow `op` at `/home/user/.local/bin/op` with a shim that transparently swaps between the primary and backup 1Password Service Account tokens on rate-limit. Real binary moves to `op-real` (mirrors `gh` → `gh-real`).

## Why

`coo-bootstrap.sh` already detects the standard SA token's 24h rate-limit at session-start and swaps to the backup token via in-process `export`. But that export only lives within the bootstrap's process tree — every subsequent Bash-tool subprocess (gh-coo-wrap's PAT resolver, direct `op read` in scripts, `schema-fetch-secrets.py`, integrity-check probes) sees the original rate-limited token again.

Live diagnosis this session: `op whoami` worked (no read cost), `op read 'op://COO/...'` returned `Too many requests`, `gh api user` fell through to `gh auth login`. The standard SA's rate-limit window has ~hours left to clear. Without the shim, the only recovery is container reboot or wait. With the shim, the alternate token absorbs the load transparently.

Diagnostic improvement after manual shim install in current session: integrity-check went 26/28 → 28/29 (S3 DEGRADED → OK because mirror-presence checks could now do op-reads; S7 SKIP → OK because credential probe could succeed). Confirms the rate-limit was masking real visibility, not just my specific gh call.

## Mechanism

1. Capture both candidate tokens as locals at shim entry.
2. Pick a "first try" by tmpfs marker preference (`$XDG_RUNTIME_DIR/coo-op-wrap/active`, contents `A` or `B`; default `A`).
3. Run `op-real`. Stdout streams through; stderr is buffered.
4. If exit non-zero AND stderr contains `rate.?limit|too many requests` AND the alternate token differs from the first → swap, retry once. Update the marker on a non-rate-limit retry outcome; leave it alone if both rate-limited.
5. Replay the final attempt's stderr only — primary RL chatter doesn't leak through on successful retry.

Install order: `session-start-sync.sh ensure_op_coo_wrap` runs before `coo-bootstrap.sh` in the SessionStart hook chain, so bootstrap's own op-reads go through the shim from boot 0. coo-bootstrap.sh's existing in-process swap also writes the marker (single new line), so the shim and bootstrap agree from the first call.

## Files

- `scripts/op-coo-wrap.sh` (new, 158 lines) — the shim itself.
- `scripts/lib/common.sh` — `ensure_op_coo_wrap()` install function, same shape as `ensure_gh_coo_wrap()`.
- `scripts/boot/session-start-sync.sh` — call it.
- `scripts/boot/coo-bootstrap.sh` — write the marker after the existing in-process swap (9 lines added).
- `scripts/ci/test-op-coo-wrap.sh` (new, 30 assertions) — covers happy path, swap-on-RL, marker-respected, swap-back when alternate RLs, both-RL fail-through, non-RL error no-swap, single-token degradation, `COO_OP_WRAP_DISABLE=1` bypass, stderr-suppression on successful retry.

## Test plan

Pre-merge (this session):

```
bash scripts/ci/test-op-coo-wrap.sh
# expect: 30 passed, 0 failed
```

Post-merge (fresh container): see handoff comment below.

## References

- `op-coo-wrap.sh` header — design notes inline.
- `coo-memory/operations/secrets/README.md §3.6` — two-token model the shim leans on.
- Phase 2 of MEMO-2026-05-22-3678 (coo-memory#873) — the secrets-scrub-from-settings.json change that made the rate-limit blast radius larger by moving op-reads to call-time.

Closes: n/a

https://claude.ai/code/session_01C3ujDgjC6263jTWSWqEZ5r

### DIFF
--- a/scripts/boot/coo-bootstrap.sh
+++ b/scripts/boot/coo-bootstrap.sh
@@ -282,6 +282,15 @@ if [ -n "${OP_SERVICE_ACCOUNT_TOKEN_BACKUP:-}" ]; then
       exit 1
     fi
     log "coo-bootstrap: BACKUP SA token active"
+    # Persist the swap for the rest of the session. The op-coo-wrap shim
+    # (installed by session-start-sync.sh ensure_op_coo_wrap, runs before
+    # this hook) reads this marker on every `op` invocation to choose the
+    # active token. Without this, every Bash-tool subprocess would re-probe
+    # the rate-limit on its first op-read. Tmpfs, container-ephemeral.
+    _op_pref_dir="${XDG_RUNTIME_DIR:-/tmp}/coo-op-wrap"
+    mkdir -p "$_op_pref_dir" 2>/dev/null || true
+    chmod 0700 "$_op_pref_dir" 2>/dev/null || true
+    printf B > "$_op_pref_dir/active" 2>/dev/null || true
     # Auto-append BACKUP's identity to the allowlist on first successful
     # swap. Authorization signal is the operator setting the BACKUP env
     # var; this avoids a FATAL on the identity-check immediately below
--- a/scripts/boot/session-start-sync.sh
+++ b/scripts/boot/session-start-sync.sh
@@ -72,6 +72,12 @@ ensure_gh_symlink_on_path
 # auto-carries the Claude Code session URL. MEMO 2026-04-26-02
 # (issue #150). Idempotent via marker grep.
 ensure_gh_coo_wrap "$SCRIPT_DIR/../gh-coo-wrap.sh"
+# Path-shadow op with op-coo-wrap so every `op` consumer gets
+# transparent rate-limit fallback between OP_SERVICE_ACCOUNT_TOKEN
+# and OP_SERVICE_ACCOUNT_TOKEN_BACKUP. Installs before
+# coo-bootstrap.sh runs (next in the SessionStart hook chain), so
+# bootstrap's own op-reads go through the shim from the start.
+ensure_op_coo_wrap "$SCRIPT_DIR/../op-coo-wrap.sh"
 # Refresh the external-touch (F6) cache when it's older than 24h.
 # Build-time prewarm in cloud-setup.sh handles the fresh-snapshot case;
 # this catches snapshots resumed after the cache has gone stale and
--- a/scripts/ci/test-op-coo-wrap.sh
+++ b/scripts/ci/test-op-coo-wrap.sh
@@ -1,0 +1,240 @@
+#!/usr/bin/env bash
+# test-op-coo-wrap: smoke-test the op wrapper that transparently
+# falls back from OP_SERVICE_ACCOUNT_TOKEN to
+# OP_SERVICE_ACCOUNT_TOKEN_BACKUP on 1Password rate-limit.
+#
+# Strategy: install a mock `op-real` whose behavior is controlled by
+# env vars (MOCK_OP_FAIL_TOKEN_A / _B set to one of {rate_limit,
+# other_error, ok}; MOCK_OP_ECHO_TOKEN_NAME=1 makes it print which
+# token bucket it received on stdout). Then run the wrapper with
+# various marker / env-var permutations and assert the swap behavior.
+#
+# Run: bash scripts/ci/test-op-coo-wrap.sh
+# Exit: 0 if all assertions pass, non-zero otherwise.
+
+set -eu
+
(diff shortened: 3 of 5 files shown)


vade-coo:
## Handoff prompt

### Pre-merge gating (this session or a `gh pr checkout 443` session)

```bash
cd "$VADE_RUNTIME_DIR"  # /home/user/coo-harness
git fetch origin claude/modest-dijkstra-P7vSK
git checkout claude/modest-dijkstra-P7vSK
bash scripts/ci/test-op-coo-wrap.sh
# expect tail: op-coo-wrap test summary: 30 passed, 0 failed; rc=0
```

Failure response: read the failing assertion + its mock trace; the test file (`scripts/ci/test-op-coo-wrap.sh`) drives a controllable mock `op-real`, so failures localize to either (a) the shim's logic for that case, or (b) the test's set-e handling around the wrapper invocation. Fix in the branch, force-push, re-run.

### Post-merge confirmation

2. `identity/charter.md`
3. `identity/governance.md`
4. `identity/preferences.md`

---

After the brake releases, in a fresh container booted from post-merge `main`:

**Step 1 — confirm the shim was installed by `session-start-sync.sh`.**

```bash
ls -la /home/user/.local/bin/op /home/user/.local/bin/op-real
grep -l COO-OP-COO-WRAP-MARKER-v1 /home/user/.local/bin/op
```

Expected: `op` is the shim (a small script with the marker), `op-real` is the 24MB ELF binary. Marker grep returns the path.

Failure response: check `/root/.vade/coo-bootstrap.log` and the `session-start-sync.sh` log fragment for `op-coo-wrap: installed`. If install was skipped, the most likely cause is `/home/user/.local/bin/op` missing at the time `ensure_op_coo_wrap` ran (cloud-setup.sh's `ensure_op_cli` is the predecessor — if it failed at build, `op` doesn't exist for the shim to wrap).

**Step 2 — confirm `op` still works through the shim.**

```bash
op whoami | head -3
op --version
```

Expected: `op whoami` returns the SA identity (URL + Integration ID + User Type lines). `op --version` returns the real op's version banner. Neither requires a rate-limit fallback — pure pass-through.

**Step 3 — confirm bootstrap wrote the marker if it swapped, OR the marker is absent if it didn't need to.**

```bash
ls -la /tmp/coo-op-wrap/ 2>/dev/null || echo "(no marker dir — expected if primary is healthy)"
cat /tmp/coo-op-wrap/active 2>/dev/null || echo "(no marker file)"
grep -E 'rate-limited|BACKUP SA token active' /root/.vade/coo-bootstrap.log | tail -5
```

Expected, primary-healthy path: no marker dir, no rate-limited lines in the bootstrap log.

Expected, rate-limited path (this only triggers when the primary's 24h window is exhausted at boot): marker file contains `B`; bootstrap log has both `standard SA token rate-limited on op read` and `BACKUP SA token active` lines.

**Step 4 — confirm the integrity check passes the S-group.**

```bash
bash "$VADE_RUNTIME_DIR/scripts/boot/integrity-check.sh" 2>&1 | tail -3
python3 -c "
import json
d = json.load(open('/home/user/.vade-cloud-state/integrity-check.json'))
s = d['groups'].get('S', {})
for k, v in sorted(s.items()):
    state = 'SKIP' if v.get('skipped') else ('OK' if v.get('ok') else 'DEGRADED')
    print(f'{k}: {state}')"
```

Expected: S3, S4, S5, S7, S8 all `OK`. S1 may still skip on PyYAML/jsonschema availability. S2 may still report unset env aliases (independent of this PR).

Failure response: if S3/S4/S5/S7 now report DEGRADED, the shim's op-read path is failing in a way the unit tests didn't catch (a flag/arg shape that the rate-limit detection doesn't pass through correctly). Surface to BDFL with the specific S-invariant's detail string.

**Step 5 — confirm `gh` works end-to-end.**

```bash
gh api user --jq .login
```

Expected: `vade-coo`. This implicitly exercises the gh-coo-wrap → op-coo-wrap chain because gh-coo-wrap's PAT cache is empty on a fresh container.

---

If all five steps pass, the change has landed cleanly. Reply with the green status; if Step 1 or 4 fails, the change should be reverted on `main` (`git revert <merge-sha>`) and re-opened as a fix PR — the install path is part of every session-start, so a broken merge takes every subsequent session offline.

https://claude.ai/code/session_01C3ujDgjC6263jTWSWqEZ5r

PR #445, opened 2026-06-05
boot: run integrity-check from coo-bootstrap exit trap

vade-coo:
## Problem

SessionStart hooks under `:startup` run **in parallel**, not sequentially.
The integrity-check call lived at the top of `coo-identity-digest.sh`
(hook #5), and it raced `coo-bootstrap.sh` (hook #3): integrity-check
observed bootstrap state mid-mutation, wrote `summary.ok=false` to the
JSON, and the digest rendered a `BOOT DEGRADED` banner warning *DO NOT
START SUBSTANTIVE WORK* — even though, by the time the agent read it,
bootstrap had already finished and the boot was green.

Recurring false alarm. Today's session is the most recent instance:
banner showed degraded (D2, D3, D4, D6, S2, S8); a fresh
`integrity-check.sh` ten seconds later reported `30/30 OK`.

## Fix

Move `integrity-check.sh` into `coo-bootstrap.sh`'s `_on_exit` trap. The
trap fires on **every** exit path — warm-skip, refuse-non-COO-gitconfig,
`CLAUDE_CODE_REMOTE!=true` skip, full-run success, and any failure — so
the check always runs after bootstrap's state mutations are durable.

Synchronization: bootstrap writes a per-session finalize sentinel at
`$HOME/.vade/.coo-bootstrap-finalized` containing `$CLAUDE_CODE_SESSION_ID`.
`coo-identity-digest.sh` polls for the sentinel with matching session ID
(60s timeout, `VADE_DIGEST_BOOTSTRAP_TIMEOUT` override) before reading
`integrity-check.json`. The match-on-session-id avoids picking up a
stale sentinel from a prior session in the same container.

## Trade-off

Bootstrap wall time grows by integrity-check's ~17s, since the two no
longer overlap in parallel. The previous overlap was the bug; this is
deliberate serialization.

## Files

- `scripts/boot/coo-bootstrap.sh` — append integrity-check + finalize
  sentinel to `_on_exit`.
- `scripts/lifecycle/coo-identity-digest.sh` — `_wait_for_bootstrap_finalized`
  at top; drop redundant integrity-check call; drop now-redundant step 3
  from the recovery banner.
- `scripts/boot/session-start-sync.sh` — comment fix (points at the new
  ordering).
- `scripts/ci/run-bootstrap-regression.sh` — comment fix (CI driver still
  invokes integrity-check explicitly because it doesn't run the full
  digest path; documented).

## Pre-merge gating

Verified in the current session by running `coo-bootstrap.sh` standalone
and confirming the trap fires:

```
$ rm -f /home/user/.vade-cloud-state/integrity-check.json \
        /root/.vade/.coo-bootstrap-finalized
$ bash $VADE_RUNTIME_DIR/scripts/boot/coo-bootstrap.sh
... coo-bootstrap: complete (~43s including integrity-check ~17s) ...
$ cat /root/.vade/.coo-bootstrap-finalized
f5936dc8-f955-4bee-b887-7e65ac184f21
$ echo $CLAUDE_CODE_SESSION_ID
f5936dc8-f955-4bee-b887-7e65ac184f21
$ jq '.summary | {ok, passed, total, degraded}' \
       /home/user/.vade-cloud-state/integrity-check.json
{"ok": true, "passed": 35, "total": 35, "degraded": []}
```

Wait function tested with matching sentinel (returns 0s) and mismatched
sentinel (times out at configured budget). Bash syntax checked on all
four edited files.

## Post-merge confirmation

The single qualifying signal is the next fresh-container boot under
the merged `main`. Handoff prompt for the next instance is posted as a
standalone PR comment per `coo-memory/operations/handoff-prompts.md`.

Refs MEMO-2026-06-04-nzxf (boot brake block-on-FAIL) — the boot-brake
flow already requires identity-stack consumption before any tool call,
so the race-fix and the brake interact cleanly.

Closes: n/a

https://claude.ai/code/session_012NhkFs7H819JByBdir8Zc3

### DIFF
--- a/scripts/boot/coo-bootstrap.sh
+++ b/scripts/boot/coo-bootstrap.sh
@@ -66,6 +66,21 @@ except Exception as e:
         ;;
     esac
   fi
+
+  # ── Hoisted integrity-check + finalize signal ─────────────────────
+  # Refresh integrity-check.json here, at the END of coo-bootstrap, so
+  # the JSON reflects FINAL post-bootstrap state instead of a parallel-
+  # hook mid-bootstrap snapshot. coo-identity-digest.sh runs in parallel
+  # under the SessionStart:startup matcher; previously its own integrity-
+  # check call observed bootstrap state mid-mutation and surfaced false-
+  # positive BOOT DEGRADED banners on otherwise-green boots. The digest
+  # now polls for the per-session finalize sentinel below before reading
+  # the JSON.
+  bash "$VADE_RUNTIME_DIR/scripts/boot/integrity-check.sh" >/dev/null 2>&1 || true
+  mkdir -p "$HOME/.vade" 2>/dev/null || true
+  printf '%s\n' "${CLAUDE_CODE_SESSION_ID:-unknown}" \
+    > "$HOME/.vade/.coo-bootstrap-finalized" 2>/dev/null || true
+
   return $rc
 }
 trap _on_exit EXIT
--- a/scripts/boot/session-start-sync.sh
+++ b/scripts/boot/session-start-sync.sh
@@ -85,7 +85,9 @@ ensure_op_coo_wrap "$SCRIPT_DIR/../op-coo-wrap.sh"
 # (no PAT at build time, etc.). Fail-open: if gh/PAT missing or refresh
 # fails, F6 falls back to its own "cache absent" skip path.
 prewarm_external_touch_cache "$WORKSPACE_ROOT_DERIVED" 24
-# integrity-check.sh runs in coo-identity-digest.sh instead of here,
-# so the check fires after the platform's repo-sync has settled
-# (coo-harness#XXX; moved from here to eliminate boot-time false alarms).
+# integrity-check.sh runs from coo-bootstrap.sh's EXIT trap, not here
+# and not from coo-identity-digest.sh. coo-bootstrap is the last hook in
+# the SessionStart:startup chain to mutate state, so deferring the check
+# to its tail eliminates the parallel-hook race that previously surfaced
+# stale BOOT DEGRADED banners on otherwise-green boots.
 boot_log_record session-start-sync end ok
--- a/scripts/ci/run-bootstrap-regression.sh
+++ b/scripts/ci/run-bootstrap-regression.sh
@@ -20,8 +20,10 @@
 #   6. Re-source the env block persisted to $TMP_HOME/.claude/settings.json
 #      (mirroring Claude Code's resume-time injection — tokens etc.),
 #      then run scripts/boot/session-start-sync.sh + integrity-check.sh
-#      explicitly (in live sessions integrity-check.sh runs inside
-#      coo-identity-digest; CI only runs session-start-sync directly).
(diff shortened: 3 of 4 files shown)


vade-coo:
## Handoff prompt — verify the fix on next fresh container

Self-contained for the next instance. Copy-paste into the next session
once `main` has the merge.

---

2. `identity/charter.md`
3. `identity/governance.md`
4. `identity/preferences.md`

### What this PR did

Moved `integrity-check.sh` from the parallel `coo-identity-digest.sh` hook into `coo-bootstrap.sh`'s `_on_exit` trap, and added a per-session finalize sentinel at `/root/.vade/.coo-bootstrap-finalized` (contains `$CLAUDE_CODE_SESSION_ID`) that the digest polls for before reading `integrity-check.json`. This eliminates the parallel-hook race that surfaced false-positive `BOOT DEGRADED` banners on otherwise-green boots.

Goal: confirm the banner reflects the **actual** boot state, not a mid-bootstrap snapshot.

**Step 1 — Read the SessionStart banner.** The digest's persisted-output should show `summary.ok=true (N/N)` and no `BOOT DEGRADED` block. If it still shows degraded, run the recovery sequence; if it shows green on the first read (most common case), the race is fixed.

**Step 2 — Confirm the finalize sentinel matches this session.**

```
$ cat /root/.vade/.coo-bootstrap-finalized
$ echo "$CLAUDE_CODE_SESSION_ID"
```

Both must print the same UUID. If the sentinel is missing or carries a different ID, the trap didn't fire for this session — investigate `coo-bootstrap.sh`'s `_on_exit` and `bootstrap_log_record` tail.

**Step 3 — Confirm `integrity-check.json` is fresh (post-bootstrap).**

```
$ stat -c '%y' /root/.vade/coo-bootstrap.log /home/user/.vade-cloud-state/integrity-check.json
$ jq '.summary | {ok, passed, total, degraded}' /home/user/.vade-cloud-state/integrity-check.json
```

JSON mtime should be **after** the bootstrap log's last `OK step=complete` line. Summary should be `ok: true`. If JSON mtime predates the bootstrap log tail, the trap's integrity-check call didn't run — check for an early exit before `_on_exit` returns.

**Step 4 — Sanity: digest wait function applied.** Grep for `_wait_for_bootstrap_finalized` in `coo-identity-digest.sh`; the function should be defined and called at the top of the script (before any read of `$INTEGRITY_CHECK`).

### Failure-mode response

If the post-merge boot shows the same false-positive degraded banner pattern (digest says degraded, a fresh `integrity-check.sh` says green seconds later), the wait function didn't block as intended. Likely causes: `CLAUDE_CODE_SESSION_ID` not exported into hook subprocesses, the sentinel-write fails silently inside the trap, or the digest's `_wait_for_bootstrap_finalized` returns before the trap fires for some exit path. Inspect `/root/.vade/coo-bootstrap.log` and `boot.log` to triage; file a follow-up PR against `coo-harness`.

https://claude.ai/code/session_012NhkFs7H819JByBdir8Zc3

PR #448, opened 2026-06-05
boot: collapse bootstrap-dependent hooks into ordered post-bootstrap chain

vade-coo:
Closes #447.

## Why

Tonight's boot exposed PR #445's wait-then-read pattern as a recurring shape, not a tunable parameter. Session `0e276dfa…`: bootstrap retry path (first attempt failed at `install_coo_ssh_keys`) plus a 36 s `integrity-check.sh` (E5–E9 live network probes) pushed total elapsed to ~78 s, past the digest's 60 s `_wait_for_bootstrap_finalized` budget. Digest read mid-write `integrity-check.json` (writes were non-atomic), node parse failed silently in `try/catch`, `_integrity_ok` stayed `"unknown"`, and the digest's branch table fell through emitting **neither** the green `Boot integrity: summary.ok=true` line **nor** the loud `BOOT DEGRADED` block. Silent on integrity state — strictly worse than the original false-positive failure mode #445 was meant to fix.

This is the fourth parallel-hook race patch in six weeks: #427 (CCR-hook ↔ git author), #442 (false-positive integrity banner), #445 (digest sentinel + 60 s wait), plus the brake's own 6 s `identity_consumed` race ([coo-memory MEMO-2026-06-04-nzxf](https://github.com/coo-labs/coo-memory/blob/main/memos/2026-06-04-nzxf.md)). Each patch addressed a symptom of one architectural choice: `SessionStart:startup` parallel execution with read-after-write dependencies between siblings. The cleanest fix removes the parallel-dependency surface entirely.

See [#447 comment](https://github.com/coo-labs/coo-harness/issues/447#issuecomment-4628606787) for the full diagnosis.

## What changed

1. **`scripts/boot/post-bootstrap-chain.sh`** — new ordered wrapper. Invokes digest → discussions-digest → project-board-digest → idle-watchdog `--start` (forked internally) → integrity-check live phase (backgrounded). Children inherit bootstrap's env directly; no sentinel handoff.
2. **`scripts/boot/coo-bootstrap.sh`** `_on_exit` trap — runs `VADE_INTEGRITY_PHASE=fast integrity-check.sh` (skips E5–E9), then `post-bootstrap-chain.sh` synchronously. Sentinel write retained for any external poller.
3. **`scripts/boot/integrity-check.sh`** — `VADE_INTEGRITY_PHASE` env gates Group E:
[...]

Issue #447, opened 2026-06-05
audit SessionStart parallel-hook reads of bootstrap-written state


vade-coo:
**Pattern:** SessionStart `:startup` hooks run in parallel, not in array order. Any hook that reads state written by another hook in the same chain races; "early" reads observe mid-mutation state and surface false signals.

**Confirmed instances (this session, 2026-06-05):**

1. `coo-identity-digest.sh`'s integrity-check call raced `coo-bootstrap.sh` and rendered a false-positive `BOOT DEGRADED` banner on an otherwise-green boot. Fixed in coo-labs/coo-harness#445 by moving integrity-check into bootstrap's `_on_exit` trap with a per-session finalize sentinel the digest polls for.
2. `discussions-digest.sh` printed `GITHUB_TOKEN unset; skipping discussions digest.` — checked its env before bootstrap populated `settings.json` env block.
3. `project-board-digest.sh` printed `GH_TOKEN unset; skipping project-board digest.` — same race shape as (2).

(2) and (3) both result in **silently skipped digests** every boot, not just degraded ones. The user-visible failure is "no In-Progress digest on the first session," but the root cause is identical to (1).

**Proposal:**

- Audit every SessionStart hook for reads of bootstrap-written state (PATs, env aliases, MCP env files, marker files, the integrity-check JSON, anything under `~/.vade/`).
- For each confirmed race, standardize on the per-session sentinel pattern coo-labs/coo-harness#445 used: a file at `$HOME/.vade/.<hook>-finalized` containing `$CLAUDE_CODE_SESSION_ID`, polled by the consumer with a bounded timeout.
- Document the pattern in `operations/` so the next hook author doesn't re-introduce a race.

**Out of scope:** any restructuring of the hook chain to actually run sequentially — that's a Claude Code surface change, not under our control. The synchronization pattern is the workable fix at our layer.

Closes: n/a

https://claude.ai/code/session_012NhkFs7H819JByBdir8Zc3

vade-coo:
## Structural alternative: collapse the chain rather than sentinel-patch each race

Tonight's post-merge confirmation of #445 surfaced a residual silent-failure that motivates a different fix than the one in this issue's Proposal. Filing this as a comment because the diagnosis and recommendation are continuous with the issue's framing, but the recommended treatment differs.

### What #445 broke open

[...]

PR #472, opened 2026-06-05
boot: observability for op-read retry waste at install_coo_ssh_keys

### LINKED ISSUE #451: install_coo_ssh_keys fails on first attempt of fresh boot, retry succeeds (~26s wasted)
## Symptom

On a fresh boot tonight (session `0e276dfa-03cf-445e-b0af-9b98d9a73d12`), the first `coo-bootstrap` attempt failed at `install_coo_ssh_keys` after 17 s, and a second attempt then ran cleanly to completion in 25 s. The bootstrap log records:

```
2026-06-05T05:04:14Z START VADE_FORCE_COO_BOOTSTRAP=0
2026-06-05T05:04:31Z FAIL step=install_coo_ssh_keys rc=1
2026-06-05T05:04:41Z START VADE_FORCE_COO_BOOTSTRAP=0
2026-06-05T05:05:06Z OK step=complete rc=0
```

The retry is what pushed the total wall time past 60 s (when combined with `integrity-check.sh`'s 36 s tail), which then expired [#445](https://github.com/coo-labs/coo-harness/pull/445)'s digest wait by ~1.2 s — that's the chain that produced the silent-integrity banner this session.

## Why this is its own issue

The structural collapse in [#448](https://github.com/coo-labs/coo-harness/pull/448) removes the race-on-slow-bootstrap shape entirely (digest is now a child, not a parallel sibling). So this `install_coo_ssh_keys` retry no longer breaks the boot banner. But the retry **still costs ~26 s of bootstrap wall time on every affected boot** — the brake's race-gate stays in PENDING for that window before all producers complete. The first-attempt failure is the root cause; the structural fix is robust to it, not a fix for it.

## Investigation candidates

- **Transient `op read` failure on first attempt.** The bootstrap's startup digest noted "standard SA token rate-limited on op read; swapping to OP_SERVICE_ACCOUNT_TOKEN_BACKUP" — possibly the same rate-limit hit the SSH key fetch on the first attempt, then the BACKUP swap let it succeed on retry. If so, the fix is to swap to BACKUP earlier (or pre-warm the read) rather than absorb the cost.
- **`ssh-keygen` fingerprint validation timing.** `install_coo_ssh_keys` validates fingerprints via `ssh-keygen`; `openssh-client` may not be installed by the time the first attempt runs (the cloud-setup snapshot warm-resume path). The digest noted "Installing openssh-client (needed for ssh-keygen fingerprint validation)" — if the install runs lazily inside the first attempt and the second attempt skips it, that explains the timing.
- **Lock / temp-file contention from a prior session.** Less likely on a fresh container but worth excluding.

## What success looks like

A fresh-boot retry rate of `0` over a 7-day window of cloud sessions, measured from `~/.vade/coo-bootstrap.log` patterns matching `FAIL step=install_coo_ssh_keys`. The bootstrap-regression CI fixture should grow a test that exercises the retry path so a regression is caught at PR time.

## Out of scope

- The race-on-slow-bootstrap shape. Fixed by [#448](https://github.com/coo-labs/coo-harness/pull/448).
- General slowness in `integrity-check.sh` E5–E9 probes. Mitigated by the `VADE_INTEGRITY_PHASE` split in [#448](https://github.com/coo-labs/coo-harness/pull/448).

## Refs

- Diagnosis session: [#448 handoff comment](https://github.com/coo-labs/coo-harness/pull/448#issuecomment-4628677604)
- Originating race issue: [#447](https://github.com/coo-labs/coo-harness/issues/447)

https://claude.ai/code/session_01NiSvz4uK2cRojqTpzVad84

vade-coo:
Closes #451.

## Why

Session `0e276dfa-…` saw `coo-bootstrap` first-attempt FAIL at `install_coo_ssh_keys` after 17 s, then succeed on a re-bootstrap 10 s later (25 s). Net waste: ~26 s of wall time per affected boot. #448's structural collapse made the banner robust to that slowness, but the retry cost stayed.

The issue body named three candidates (transient `op read`, lazy openssh-client install, lock contention) without a confirmed root cause — the failing session's actual stderr was never captured. Pre-recon ruled out candidates 2 and 3 (the prod base image already has `openssh-client`; fresh container = no contention). Candidate 1 remains a working hypothesis, but the op-coo-wrap shim only swaps tokens on rate-limit-shaped stderr; a non-rate-limit transient (503, TLS cold-start) would burn the full retry budget without triggering the swap. Same wall-clock cost, hidden cause.

This PR is the **observability + CI-fixture half** of the fix. The structural fix (pre-warm BACKUP swap, widen the rate-limit regex, or another data-driven move) is deferred until the new jsonl surfaces the actual error shape from a real failure.

## What changed

1. **`scripts/lib/common.sh`** — `_op_to_file` captures stderr from the inner retry chain, replays it to the operator (preserving existing log behavior), and writes a structured event to `~/.vade/op-read-failures.jsonl` when the call failed (rc!=0) or burned >1 s wall time (i.e., ≥1 retry fired). Schema:
   ```json
   {"ts":"…","kind":"op_read","ref":"op://…","rc":0,"elapsed_ms":3022,
    "op_wrap_pref_before":"A","op_wrap_pref_after":"B","stderr_tail":"…"}
   ```
   Compact-separator JSON for jsonl convention + grep-friendliness. New helper `_record_op_read_event` does the python3 write so bash escaping doesn't bite us. Successful sub-1 s reads are not logged (too noisy).

2. **`scripts/boot/coo-bootstrap.sh`** — exports `COO_OP_WRAP_TRACE=1` right after the START log line. The shim writes one line per op-invocation decision (which token, swap or not, rate-limit detected) to `/dev/shm/coo-op-wrap.trace`. Pairs with the jsonl event to give the wrapper-side picture. Container-ephemeral.

3. **`scripts/boot/integrity-check.sh`** — new **D7** probe. Counts events in the jsonl over a rolling window (default 24 h, overridable via `VADE_D7_WINDOW_H`). `true` when zero events or only slow-but-recoverable; `false` when any event has rc!=0. Detail line cites the path so triage is one `cat` away. This is the metric for #451's success criterion ("zero retries over 7-day window") — it gates the eventual close.

4. **`scripts/ci/mocks/op`** — `VADE_CI_MOCK_OP_FAIL_N` env makes the next N `op read` invocations exit 1 with rate-limit stderr. Counter file in `/tmp`; only affects `read`, so the bootstrap's non-retry-bounded `op whoami` / `op run` calls aren't masked.

5. **`scripts/ci/test-op-retry-recovery.sh`** — new standalone test, modeled on `test-op-coo-wrap.sh`. Overrides `op` as a bash function with a counter and asserts:
   - **N=2** → `_op_to_file` recovers within the 5-retry budget (rc=0, target file written, exactly 3 attempts consumed, jsonl event lands with rc=0).
   - **N=5** → all retries fail (rc=1, no half-written target, full 5-attempt budget consumed, jsonl event with rc!=0 lands).
   
   8 assertions, ~16 s runtime (dominated by retry backoff sleeps).

6. **`.github/workflows/bootstrap-regression.yml`** — wires the new test into the `transcript-redaction` job alongside the other `test-*.sh` steps. Filed via the `vade-coo-app` install-token contents API because the session OAuth lacks `workflow` scope (commit `410534fc` authored by `vade-coo-app[bot]`).

## Verification

In-session:
- Bash syntax: clean on all five touched files (`bash -n`).
- New test: 8 passed, 0 failed.
- Adjacent suites unaffected: `test-op-coo-wrap.sh` 30/30, `test-bootstrap-op-caches.sh` 17/17.
- `integrity-check.sh` D7: lands `true` on this session ("no op-read-failures.jsonl"). Synthetic failure event flips to `false` with the path-cited detail line. Group D goes from 6 to 7 invariants; full check 34/34.

Live container behavior requires a fresh boot. **Handoff prompt posted as a PR comment per CLAUDE.md.**

## Out of scope (separate work, will file)

- **Speculative structural fix** (pre-warm BACKUP swap before `install_coo_ssh_keys` or widen `_is_rate_limit`). Deferred until the new jsonl surfaces a real failure's error shape — fixing-then-measuring would be guess-driven.
- **Idempotent `install_coo_ssh_keys`** — re-runs op-read every boot even when the four key files already exist on disk with valid fingerprints. Tangential to #451 (doesn't help fresh-boot first attempt) but real adjacent waste. Filing as separate area:bootstrap issue.

## Linked work

- Issue: #451
- Structural precursor (out-of-scope race fix): #448
- Originating race issue: #447

https://claude.ai/code/session_01RuoFyvVGb6dcqp7jJi3PMH

### DIFF
--- a/.github/workflows/bootstrap-regression.yml
+++ b/.github/workflows/bootstrap-regression.yml
@@ -134,6 +134,9 @@ jobs:
       - name: Auto-subscribe PR hook — matcher contract (wrapper + raw)
         run: bash "$GITHUB_WORKSPACE/scripts/ci/test-auto-subscribe-pr.sh"
 
+      - name: op-read retry recovery — install_coo_ssh_keys path (coo-harness#451)
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/test-op-retry-recovery.sh"
+
       - name: Git push fallback — op:// PAT resolution + leak hardening (coo-harness#457, #124)
         run: bash "$GITHUB_WORKSPACE/scripts/ci/test-git-push-fallback.sh"
 
--- a/scripts/boot/coo-bootstrap.sh
+++ b/scripts/boot/coo-bootstrap.sh
@@ -227,6 +227,15 @@ fi
 log "coo-bootstrap: starting"
 bootstrap_log_record START "VADE_FORCE_COO_BOOTSTRAP=${VADE_FORCE_COO_BOOTSTRAP:-0}"
 
+# Enable op-coo-wrap decision tracing for the bootstrap process and any
+# child processes that inherit env. The shim writes one line per op-read
+# decision (which token, swap or not, rate-limit detected) to
+# /dev/shm/coo-op-wrap.trace. Pairs with _op_to_file's per-call jsonl
+# event in ~/.vade/op-read-failures.jsonl to surface the actual error
+# shape behind install_coo_ssh_keys first-attempt FAIL (coo-harness#451).
+# Container-ephemeral; zero cost when the file isn't written.
+export COO_OP_WRAP_TRACE=1
+
 COO_BOOTSTRAP_STEP="ensure_op_cli"
 ensure_op_cli
 
--- a/scripts/boot/integrity-check.sh
+++ b/scripts/boot/integrity-check.sh
@@ -370,6 +370,58 @@ else
   _add D6 false "vade-coo-{auth,sign} keys missing in $HOME/.ssh"
 fi
 
+# D7 — op-read failure surface (coo-harness#451)
+# Counts events in ~/.vade/op-read-failures.jsonl over a rolling window.
+# Each event is one _op_to_file call that either failed outright or took
+# >1s wall time (i.e., burned at least one retry). The metric for #451's
+# success criterion ("zero retries over 7-day window") rolls up from
+# this file across the cloud-session fleet.
+#
+# Probe is `true` when:
+#   - no events in window (clean), or
+#   - events present but rc=0 across the board (slow but recoverable);
+#     surfaced as a non-blocking detail so operators can see the cost.
+# Probe is `false` when any event has rc!=0 — that's an
+# install_coo_ssh_keys-class FAIL that wasted bootstrap wall time and
+# may have triggered a retry-bootstrap cycle.
+D7_log="${HOME}/.vade/op-read-failures.jsonl"
+D7_window_h="${VADE_D7_WINDOW_H:-24}"
+if [ -f "$D7_log" ] && check_cmd python3; then
+  D7_summary="$(python3 -c "
+import json, datetime
+window_h = ${D7_window_h}
+cutoff = datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(hours=window_h)
+total = 0; fails = 0; total_ms = 0
+try:
+    with open('$D7_log') as f:
+        for line in f:
+            try: ev = json.loads(line)
+            except Exception: continue
+            try: t = datetime.datetime.strptime(ev.get('ts',''),'%Y-%m-%dT%H:%M:%SZ').replace(tzinfo=datetime.timezone.utc)
+            except Exception: continue
+            if t < cutoff: continue
+            total += 1
+            if ev.get('rc') != 0: fails += 1
+            try: total_ms += int(ev.get('elapsed_ms',0) or 0)
+            except Exception: pass
+except Exception: pass
+print(f'{total}|{fails}|{total_ms}')
+" 2>/dev/null || echo "0|0|0")"
+  D7_total="${D7_summary%%|*}"
+  _D7_rest="${D7_summary#*|}"
+  D7_fails="${_D7_rest%%|*}"
+  D7_ms="${_D7_rest#*|}"
+  if [ "$D7_total" = "0" ]; then
+    _add D7 true "no op-read events in last ${D7_window_h}h"
+  elif [ "$D7_fails" = "0" ]; then
+    _add D7 true "${D7_total} slow op-read event(s) in last ${D7_window_h}h, 0 failed (${D7_ms}ms cumulative); see $D7_log"
+  else
+    _add D7 false "${D7_fails}/${D7_total} op-read event(s) failed in last ${D7_window_h}h (${D7_ms}ms cumulative wall-time cost); see $D7_log"
+  fi
+else
+  _add D7 true "no op-read-failures.jsonl (clean steady state)"
+fi
+
 # ── Group E: MCP surface ─────────────────────────────────────
 # Phase gate. `fast` (the boot-path call from coo-bootstrap's exit
 # trap) marks E1–E9 as `pending` and skips the live probes entirely,
--- a/scripts/ci/mocks/op
+++ b/scripts/ci/mocks/op
@@ -19,6 +19,28 @@ set -euo pipefail
 
 KEYDIR="${VADE_CI_MOCK_KEYDIR:-/tmp/vade-ci-mocks/keys}"
 
+# Failure injection for the retry-recovery test (coo-harness#451).
+# When VADE_CI_MOCK_OP_FAIL_N is a positive integer, the next N `op read`
+# invocations return non-zero with rate-limit-shaped stderr. State lives
+# in a counter file so it survives across subshells inside a single CI
+# run; the file is recreated per test invocation. Subcommands other than
+# `read` are not affected (the bootstrap calls `op whoami` and `op run`
+# without retry budgets — failing those would mask the test's intent).
+if [ "${1:-}" = "read" ] \
+   && [ -n "${VADE_CI_MOCK_OP_FAIL_N:-}" ] \
+   && [ "${VADE_CI_MOCK_OP_FAIL_N}" -gt 0 ] 2>/dev/null; then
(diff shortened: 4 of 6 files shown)


vade-coo:
## Handoff prompt — verify on next fresh container

Self-contained for the next instance. Copy-paste into the next session on a fresh cloud container with this PR's branch merged (or use the PR's branch directly via `gh pr checkout 472`).

---

Per [MEMO-2026-06-04-nzxf](https://github.com/coo-labs/coo-memory/blob/main/memos/2026-06-04-nzxf.md) the boot brake is `block-on-FAIL`. It refuses every non-Read/non-Grep tool call until you have Read all four identity-stack files and the persisted-output gate file (if the digest banner declared one). **Do not skip.** Follow CLAUDE.md §"Session-start mandatory reading order" verbatim:

2. `identity/charter.md`
3. `identity/governance.md`
4. `identity/preferences.md`

One parallel batch is fine. Once the brake releases, the verification commands below will work.

### What this PR did

Observability + CI-fixture half of the [#451](https://github.com/coo-labs/coo-harness/issues/451) fix. **No structural change to the bootstrap flow on the success path.** Adds:

- `_op_to_file` writes a jsonl event to `~/.vade/op-read-failures.jsonl` when an op-read call fails or burns >1 s.
- `coo-bootstrap.sh` exports `COO_OP_WRAP_TRACE=1` so the shim writes per-invocation decisions to `/dev/shm/coo-op-wrap.trace`.
- `integrity-check.sh` D7 probe surfaces failure count over a rolling 24 h window.
- CI mock op fail-injection (`VADE_CI_MOCK_OP_FAIL_N`) + `test-op-retry-recovery.sh` lock in the recovery behavior at PR time.

Structural fix (pre-warm BACKUP swap or widen `_is_rate_limit`) is deferred — the next failure that surfaces will tell us which shape we're fighting.

**Step 1 — Banner shape unchanged.** Boot integrity line should still read `summary.ok=true (N/N)` with N=34 or 35 (depends on environment). Group D goes from 6 → 7 invariants. D7 should show `true: no op-read events in last 24h` on a clean cold boot.

```
$ jq '.groups.D.D7' "$HOME/.vade-cloud-state/integrity-check.json"
```

Expect `{"ok": true, "detail": "no op-read-failures.jsonl …" OR "no op-read events in last 24h"}`.

**Step 2 — Wrapper trace fires.** During bootstrap, the op-coo-wrap shim should write to `/dev/shm/coo-op-wrap.trace` on every `op` invocation:

```
$ ls -la /dev/shm/coo-op-wrap.trace
$ head -10 /dev/shm/coo-op-wrap.trace
```

Expect lines like `2026-06-05T… attempt 1: token=A argv=read op://COO/vade-coo-ssh-auth/private key`. Empty file → `COO_OP_WRAP_TRACE=1` did not reach the shim's env (regression).

**Step 3 — jsonl writes on first transient (if any).** Cold cloud boots sometimes burn a retry on transient op-read failures. If so, the new path should record it:

```
$ cat "$HOME/.vade/op-read-failures.jsonl" 2>/dev/null
```

Expect either no file (clean boot, no retries triggered) or jsonl lines like:
```json
{"ts":"2026-06-05T…","kind":"op_read","ref":"op://COO/vade-coo-ssh-auth/private key","rc":0,"elapsed_ms":3142,"op_wrap_pref_before":"A","op_wrap_pref_after":"B","stderr_tail":"…"}
```

If a line has `"rc":1`, D7 should be `false` on that session — confirming the surface fires end-to-end.

**Step 4 — install_coo_ssh_keys success path unaffected.** Bootstrap log should show the canonical clean tail:

```
$ tail -3 "$HOME/.vade/coo-bootstrap.log"
```

Expect `OK step=complete rc=0` on the most recent line. The `Installing COO SSH keys into …` / `wrote …` / `SSH key fingerprints validated` log lines should appear in the SessionStart output as before.

**Step 5 — CI fixture caught.** The bootstrap-regression workflow should report a new step *"op-read retry recovery — install_coo_ssh_keys path (coo-harness#451)"* with rc=0 in the transcript-redaction job. The test takes ~16 s.

### Failure-mode response

- **D7 is `false` on first cold boot** — the new jsonl just caught an actual transient. **This is the success case for observability.** Don't panic. Read the failing event's `stderr_tail` and `op_wrap_pref_before/after` — that's the data point we've been missing. File a follow-up referencing #451 with the captured stderr, and we can now make the data-driven structural-fix call.
- **`/dev/shm/coo-op-wrap.trace` empty or missing** — `COO_OP_WRAP_TRACE=1` didn't propagate to the shim's env. Check `coo-bootstrap.sh` line ~233 for the export. Regression of the PR's change.
- **`~/.vade/op-read-failures.jsonl` writes on a successful sub-1 s op-read** — the elapsed-ms gate (line ~2049 in `common.sh`) is misbehaving. Probably the `date +%s%3N` substitution returning a stale value or a system without `%3N` support. Surface, don't block.
- **D7 detail line cites the wrong path** — string-substitution bug in the python heredoc. Inspect `integrity-check.sh:D7_summary`.

### Companion follow-up

Once the jsonl captures a real failure shape (likely within days based on the issue's "fourth race in six weeks" cadence), the data-driven structural fix lands as a separate PR. Candidates pre-recon: pre-warm BACKUP swap in `coo-bootstrap.sh` before `install_coo_ssh_keys`, or widen `op-coo-wrap.sh:_is_rate_limit` to cover 503 / "context deadline" shapes if those turn out to be the actual transient.

Separate but adjacent: idempotent `install_coo_ssh_keys` (skip op-reads when disk keys + fingerprints already match expected). Tangential to #451 — filed separately under area:bootstrap, M2.

https://claude.ai/code/session_01RuoFyvVGb6dcqp7jJi3PMH

PR #475, opened 2026-06-05
boot: idempotent install_coo_ssh_keys — skip op-read when disk matches

### LINKED ISSUE #473: install_coo_ssh_keys: skip op-reads when disk keys + fingerprints already match
## Symptom

`install_coo_ssh_keys` (`scripts/lib/common.sh:1961`) unconditionally op-reads the four SSH key fields on every bootstrap invocation:

```
op://COO/vade-coo-ssh-auth/private key
op://COO/vade-coo-ssh-auth/public key
op://COO/vade-coo-ssh-sign/private key
op://COO/vade-coo-ssh-sign/public key
```

It writes them to `~/.ssh/`, then validates fingerprints against `COO_AUTH_FP_EXPECTED` / `COO_SIGN_FP_EXPECTED`. There is no presence-and-validity check that short-circuits the op-reads when the disk state is already correct.

## Impact

Observed this session: the bootstrap ran twice — first cold, then a re-bootstrap triggered by "marker present but cached `GITHUB_MCP_PAT` no longer authenticates as vade-coo" (`coo-bootstrap.sh:218`). The second run re-issued all four `op read` calls and re-wrote the four key files even though the disk state was already identical to the op:// content. Each `op read` counts against the daily SA quota and burns wall time.

Quantified cost per bootstrap re-run: 4 `op read` calls × ~0.4 s typical latency = ~1.6 s wall time, plus 4 quota-bucket consumption against a 1000 read+write/24h budget (Personal/Teams). Stale-PAT re-bootstraps are recurring (the marker-stale branch is the common path on multi-snapshot resume); the cost compounds.

## Why this is a separate issue

This does **not** address [#451](https://github.com/coo-labs/coo-harness/issues/451) — the fresh-boot first-attempt FAIL pattern. On a fresh container, the four key files don't exist on disk yet, so an idempotent skip wouldn't fire on the failing path. The two issues touch the same function for different reasons:

- #451: first-attempt FAIL costs ~17 s + a full re-bootstrap (~25 s).
- This: re-bootstrap on already-installed keys costs ~1.6 s + 4 quota units.

Conflating them in one PR muddles the success metric for each. #451's metric is failure-rate over a 7-day window; this issue's metric is bootstrap-runtime + quota savings on stale-PAT re-runs.

## Proposal

Add a presence-and-validity gate at the top of `install_coo_ssh_keys` before the four `_op_to_file` calls:

```sh
if [ -f "$ssh_dir/vade-coo-auth" ] \
   && [ -f "$ssh_dir/vade-coo-auth.pub" ] \
   && [ -f "$ssh_dir/vade-coo-sign" ] \
   && [ -f "$ssh_dir/vade-coo-sign.pub" ] \
   && ensure_openssh_client; then
  local fp_a fp_s
  fp_a="$(_fingerprint_of "$ssh_dir/vade-coo-auth.pub")"
  fp_s="$(_fingerprint_of "$ssh_dir/vade-coo-sign.pub")"
  if [ "$fp_a" = "$COO_AUTH_FP_EXPECTED" ] && [ "$fp_s" = "$COO_SIGN_FP_EXPECTED" ]; then
    log "COO SSH keys already installed (fingerprints match); skipping op-read"
    # Still ensure allowed_signers + known_hosts even on the skip path
    # since those are derived state, not the keys themselves.
    return 0
  fi
  log "COO SSH keys present but fingerprint mismatch (auth=$fp_a sign=$fp_s); re-fetching"
fi
```

Fail-open semantics:
- Missing `ssh-keygen` → fall through to the existing fetch path (fingerprint check would have been skipped anyway).
- Fingerprint mismatch on disk → fall through to fetch (overwrites with current op:// content).
- Either pubkey missing → fall through.

The op-read fetch path stays as the canonical write-path for any non-clean disk state. The skip path is a pure-presence optimization.

## Decision rationale (for case-law)

This is the "trust disk + validate fingerprints, refetch on mismatch" pattern, not "trust disk blindly." The current code is conservative-to-a-fault — it pays 4 op-reads per boot to avoid trusting disk state at all. Once we have fingerprint validation (`ensure_openssh_client` already on path), trusting the disk is well-bounded.

## What success looks like

- Re-bootstrap (`marker stale` or `force=1` re-run) completes `install_coo_ssh_keys` in <100 ms when disk state is already correct.
- 4-credential reduction in op-read count per re-bootstrap, surfacing as lower D7 traffic in `~/.vade/op-read-failures.jsonl` (only the >1 s call path logs there — but the wrapper trace at `/dev/shm/coo-op-wrap.trace` will show 4 fewer entries per re-boot).
- CI fixture in `scripts/ci/test-op-retry-recovery.sh` style: stage a fixture `$HOME/.ssh` with valid keys, run install_coo_ssh_keys with a mocked-to-fail op, assert it returns 0 and never called op.

## Out of scope

- The fresh-boot first-attempt FAIL path ([#451](https://github.com/coo-labs/coo-harness/issues/451)).
- Cache invalidation policy when 1Password key rotates — the fingerprint mismatch path handles that (mismatch → refetch).

## Refs

- Affected function: [`scripts/lib/common.sh:1961`](https://github.com/coo-labs/coo-harness/blob/main/scripts/lib/common.sh#L1961) (`install_coo_ssh_keys`)
- Stale-PAT re-bootstrap path: [`scripts/boot/coo-bootstrap.sh:218`](https://github.com/coo-labs/coo-harness/blob/main/scripts/boot/coo-bootstrap.sh#L218)
- Parent observability PR: [#472](https://github.com/coo-labs/coo-harness/pull/472)

https://claude.ai/code/session_01RuoFyvVGb6dcqp7jJi3PMH

vade-coo:
Closes #473.

## Why

`install_coo_ssh_keys` issues four 1Password reads every bootstrap. On a fresh boot that's load-bearing; on a marker-stale re-bootstrap it's pure waste — disk already holds the correct keys from the cold boot 30 s ago, and the four `op read`s just rewrite identical bytes. Each one is ~0.4 s wall time + one quota unit against the per-SA daily budget (1000 read+write/24h on Personal/Teams plans).

Our own session today shows the shape:
```
09:58:25Z START
09:58:55Z OK step=complete rc=0     (30s cold)
09:59:26Z START marker stale ...
09:59:35Z OK step=complete rc=0     (9s warm re-run — but four redundant op-reads inside)
```

Separate from [#451](https://github.com/coo-labs/coo-harness/issues/451) (fresh-boot first-attempt FAIL). #451 is about recovery from transients on a no-disk-state boot; #473 is about not paying for a fetch when disk is already authoritative. The two issues touch the same function for orthogonal reasons.

## What changed

**`scripts/lib/common.sh` `install_coo_ssh_keys`** — new skip-when-clean gate at the top:

```
if four-key-files-exist && ssh-keygen-available && both-fingerprints-match-expected; then
    log "COO SSH keys already installed at $ssh_dir (fingerprints match); skipping op-read"
    refresh allowed_signers + known_hosts
    return 0
fi
# else: fall through to the existing op-read + fingerprint-check + write path
```

Truth table:

| keys-on-disk | ssh-keygen | fp match | path           |
|---|---|---|---|
| any-missing  | -          | -        | full fetch     |
| all-present  | absent     | -        | full fetch (can't validate → safer to refetch) |
| all-present  | present    | mismatch | full fetch (logs `refetching` — stale or rotated key) |
| all-present  | present    | match    | **skip** (logs `skipping op-read`) |

The `allowed_signers` / `ssh-keyscan` derived-state section is extracted into `_install_coo_ssh_keys_derived_state` so both paths reach identical disk state without code duplication.

## Test

New `scripts/ci/test-install-coo-ssh-keys-idempotent.sh`, modeled on `test-op-retry-recovery.sh`. Generates fixture ed25519 keys, captures their fingerprints as `COO_*_FP_EXPECTED`, then asserts:

1. **Skip path** — disk pre-staged with matching fingerprints; `op` is never called (0 invocations); `rc=0`; skip-line logged; `allowed_signers` refreshed.
2. **Mismatch path** — flip the expected auth fingerprint; `op` IS called (4 invocations); `refetching` line logged; `rc=1` (post-fetch check rejects against the now-mismatched expected — confirming the AND-of-both gate, not OR).
3. **Missing-file path** — `rm` one key file before invocation; the gate drops to the fetch branch even though three files are present; `op` called ≥4 times; `rc=0` (mock writes content matching the reset expected fingerprints).

10 assertions, runs in <1 s (no retry sleeps). Wired into the `transcript-redaction` job in `.github/workflows/bootstrap-regression.yml`.

## Verification in-session

- `bash -n scripts/lib/common.sh` clean.
- New test: 10/10 passing.
- Adjacent suites unaffected: `test-op-retry-recovery.sh` 8/8, `test-op-coo-wrap.sh` 30/30.

Live container behavior requires a fresh boot. Handoff prompt posted as PR comment per CLAUDE.md "Handoff prompts for boot-impacting PRs".

## Linked work

- Issue: #473
- Sibling (orthogonal but adjacent): #451 / [#472](https://github.com/coo-labs/coo-harness/pull/472)

https://claude.ai/code/session_01RuoFyvVGb6dcqp7jJi3PMH

### DIFF
--- a/.github/workflows/bootstrap-regression.yml
+++ b/.github/workflows/bootstrap-regression.yml
@@ -137,6 +137,9 @@ jobs:
       - name: op-read retry recovery — install_coo_ssh_keys path (coo-harness#451)
         run: bash "$GITHUB_WORKSPACE/scripts/ci/test-op-retry-recovery.sh"
 
+      - name: install_coo_ssh_keys idempotent skip (coo-harness#473)
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/test-install-coo-ssh-keys-idempotent.sh"
+
       - name: Git push fallback — op:// PAT resolution + leak hardening (coo-harness#457, #124)
         run: bash "$GITHUB_WORKSPACE/scripts/ci/test-git-push-fallback.sh"
 
--- a/scripts/ci/test-install-coo-ssh-keys-idempotent.sh
+++ b/scripts/ci/test-install-coo-ssh-keys-idempotent.sh
@@ -1,0 +1,206 @@
+#!/usr/bin/env bash
+# test-install-coo-ssh-keys-idempotent: lock in the skip-when-clean
+# behavior of install_coo_ssh_keys (coo-harness#473).
+#
+# When the four COO SSH key files already exist on disk and their
+# fingerprints match the expected values, install_coo_ssh_keys should
+# not call `op read` at all — the disk content would be identical, and
+# every avoided op-read saves wall time + 1P quota.
+#
+# Strategy: source common.sh, override `op` as a bash function that
+# tracks whether it was called. Pre-stage $HOME/.ssh with fixture
+# ed25519 keys; set COO_AUTH_FP_EXPECTED / COO_SIGN_FP_EXPECTED to the
+# matching values. Run install_coo_ssh_keys and assert op was never
+# invoked. Then mutate one expected fingerprint and assert the function
+# falls through to the op path.
+#
+# Run: bash scripts/ci/test-install-coo-ssh-keys-idempotent.sh
+# Exit: 0 if all assertions pass, non-zero otherwise.
+
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+COMMON="$REPO_ROOT/scripts/lib/common.sh"
+
+[ -f "$COMMON" ] || { echo "FAIL: common.sh not found at $COMMON" >&2; exit 1; }
+
+command -v ssh-keygen >/dev/null 2>&1 || {
+  echo "SKIP: ssh-keygen not on PATH; idempotent skip relies on it" >&2
+  exit 0
+}
+
+WORK="$(mktemp -d)"
+trap 'rm -rf "$WORK"' EXIT
+
+export HOME="$WORK/home"
+mkdir -p "$HOME/.ssh" "$HOME/.vade"
+chmod 700 "$HOME/.ssh"
+
+export VADE_RUNTIME_DIR="$REPO_ROOT"
+export VADE_COO_MEMORY_DIR="${VADE_COO_MEMORY_DIR:-$REPO_ROOT/../coo-memory}"
+export VADE_CLOUD_STATE_DIR="$WORK/cloud-state"
+mkdir -p "$VADE_CLOUD_STATE_DIR"
+
+# Generate fixture ed25519 keys; capture their fingerprints as the
+# "expected" values for the skip path. Same pattern the bootstrap-
+# regression runner uses (scripts/ci/run-bootstrap-regression.sh).
+ssh-keygen -t ed25519 -N "" -C "test-auth" -f "$HOME/.ssh/vade-coo-auth"  >/dev/null
+ssh-keygen -t ed25519 -N "" -C "test-sign" -f "$HOME/.ssh/vade-coo-sign"  >/dev/null
+chmod 0600 "$HOME/.ssh/vade-coo-auth" "$HOME/.ssh/vade-coo-sign"
+chmod 0644 "$HOME/.ssh/vade-coo-auth.pub" "$HOME/.ssh/vade-coo-sign.pub"
+
+COO_AUTH_FP_EXPECTED="$(ssh-keygen -lf "$HOME/.ssh/vade-coo-auth.pub" | awk '{print $2}')"
+COO_SIGN_FP_EXPECTED="$(ssh-keygen -lf "$HOME/.ssh/vade-coo-sign.pub" | awk '{print $2}')"
+export COO_AUTH_FP_EXPECTED COO_SIGN_FP_EXPECTED
+
+# shellcheck disable=SC1090
+source "$COMMON"
+
+PASS=0
+FAIL=0
+declare -a FAILURES=()
+
+_pass() { PASS=$((PASS+1)); printf '  PASS  %s\n' "$1"; }
+_fail() { FAIL=$((FAIL+1)); FAILURES+=("$1"); printf '  FAIL  %s\n' "$1"; }
+
+OP_CALL_LOG="$WORK/op-calls.log"
+: > "$OP_CALL_LOG"
+
+# Snapshot the on-disk key material at setup so the mock can return
+# stable content even after test 3 deletes one of the files. Without
+# this, the mock's `cat` would return empty content on the deleted
+# file, _op_to_file's empty-content guard would trip, and the test
+# would fail for the wrong reason (mock plumbing, not the skip-branch
+# fall-through we're checking).
+SNAP_AUTH_PRIV="$(cat "$HOME/.ssh/vade-coo-auth")"
+SNAP_AUTH_PUB="$(cat "$HOME/.ssh/vade-coo-auth.pub")"
+SNAP_SIGN_PRIV="$(cat "$HOME/.ssh/vade-coo-sign")"
+SNAP_SIGN_PUB="$(cat "$HOME/.ssh/vade-coo-sign.pub")"
+export SNAP_AUTH_PRIV SNAP_AUTH_PUB SNAP_SIGN_PRIV SNAP_SIGN_PUB
+
+# Tracking op stub — records each invocation. The success branch echoes
+# the snapshotted content so _op_to_file's empty-content guard doesn't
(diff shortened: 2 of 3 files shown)


vade-coo:
## Handoff prompt — verify on next fresh container

Self-contained for the next instance. Copy-paste into the next session on a fresh cloud container with this PR's branch merged.

---

Per [MEMO-2026-06-04-nzxf](https://github.com/coo-labs/coo-memory/blob/main/memos/2026-06-04-nzxf.md) the boot brake is `block-on-FAIL`. It refuses every non-Read/non-Grep tool call until you have Read all four identity-stack files and the persisted-output gate file (if the digest banner declared one). Follow CLAUDE.md §"Session-start mandatory reading order" verbatim.

### What this PR did

Adds a skip-when-clean gate to `install_coo_ssh_keys` (`scripts/lib/common.sh`). When all four COO SSH key files exist on disk AND `ssh-keygen` is available AND both pubkey fingerprints match `COO_AUTH_FP_EXPECTED` / `COO_SIGN_FP_EXPECTED`, the four `op read` calls are skipped — they would just rewrite identical bytes. Saves ~1.6 s + 4 quota units per marker-stale re-bootstrap.

**Step 1 — Cold boot path unchanged.** First boot after container creation has no keys on disk. install_coo_ssh_keys must take the full-fetch path:

```
$ grep -E 'Installing COO SSH keys|skipping op-read' /tmp/claude-*/-home-user/*/tool-results/hook-*-stdout.txt 2>/dev/null | head -5
```

Expect `Installing COO SSH keys into /home/user/.ssh` (the fetch-path log line), NOT `skipping op-read`.

```
$ ls -la "$HOME/.ssh/vade-coo-"*
```

Expect four files with the expected sizes / fingerprints.

**Step 2 — Re-bootstrap takes the skip path.** Triggered by `marker stale` after the cold boot. Look for the skip log line in the SECOND coo-bootstrap entry of the session:

```
$ tail -n 30 "$HOME/.vade/coo-bootstrap.log"
```

Expect two `OK step=complete rc=0` entries. The second entry corresponds to a bootstrap run that should NOT have called op-read on the SSH key fields. Confirm via the op-coo-wrap trace (enabled in coo-harness#472):

```
$ grep -c 'argv=read op://COO/vade-coo-ssh-' /dev/shm/coo-op-wrap.trace
```

This counts op-read invocations against ssh-key refs across the whole session. Expect **exactly 4** (the cold boot's four), not 8 (cold + warm).

**Step 3 — Skip-path keeps derived state correct.** The skip path still refreshes `allowed_signers` and best-effort `ssh-keyscan`:

```
$ stat -c '%y %n' "$HOME/.ssh/allowed_signers"
$ head -2 "$HOME/.ssh/allowed_signers"
```

Expect mtime within a couple of seconds of the warm re-bootstrap (proving the skip path ran the derived-state helper), and two `coo@vade-app.dev ssh-ed25519 …` lines.

**Step 4 — CI fixture caught.** `bootstrap-regression` workflow should show two new test steps under `transcript-redaction`:
- *"op-read retry recovery — install_coo_ssh_keys path (coo-harness#451)"* (from #472)
- *"install_coo_ssh_keys idempotent skip (coo-harness#473)"* (this PR)

Both green, sub-second each (no retry sleeps).

### Failure-mode response

- **Cold boot takes the skip path (skip-line logged on first install_coo_ssh_keys call)** — gate misbehaving on a no-keys-on-disk container. Likely a stat-cache or remnant from a snapshot. Check the four file paths and the marker file presence. File a follow-up.
- **Re-bootstrap takes the full-fetch path (no skip-line, op-coo-wrap trace shows 8 ssh-key reads instead of 4)** — gate's fingerprint check is rejecting valid keys. Likely a fingerprint format mismatch between `_fingerprint_of` (awk-parsed) and the expected env var. Read `coo-bootstrap.log` for the `fingerprint mismatch (auth=… sign=…); refetching` line — the captured values tell you which one diverged.
- **Skip path triggers but `allowed_signers` not refreshed** — `_install_coo_ssh_keys_derived_state` failed silently. Read the bootstrap stderr for `Wrote …/allowed_signers` line.
- **Cold boot fingerprint check fails after fetch** — unrelated to this PR; the upstream `COO_AUTH_FP_EXPECTED` literal in `scripts/lib/common.sh:707` is stale (1Password key rotated). Rotate-track issue, not a #473 regression.

### Companion

Pairs with [#472](https://github.com/coo-labs/coo-harness/pull/472) (observability — first half of [#451](https://github.com/coo-labs/coo-harness/issues/451)). When the next fresh-boot first-attempt FAIL surfaces in `~/.vade/op-read-failures.jsonl`, the data-driven structural fix lands as a separate PR.

https://claude.ai/code/session_01RuoFyvVGb6dcqp7jJi3PMH

